PR #89, opened 2014-12-16
Removed underscore properties in annotations.

### LINKED ISSUE #75: Variable names does not follow ongr-strict-standard
Member variable "index_analyzer" is not in valid camel caps format
Member variable "saerch_analyzer" is not in valid camel caps format

ElasticsearchBundle/Annotation/MultiField.php
ElasticsearchBundle/Annotation/Property.php

Zylius:
Closes #75.
> Note that I don't like this myself, but I wasn't able to figure out a better way to do this. In my opinion, annotation setting names and elasticsearch setting names should match, so it's easier to understand for end user. Instead of converting this from underscore to camelCase and back to underscore again, I removed the properties and store them in array. Please note that we would have to use annotation to define properties and constructor either way.

### DIFF
--- a/Annotation/MultiField.php
+++ b/Annotation/MultiField.php
@@ -16,42 +16,43 @@ namespace ONGR\ElasticsearchBundle\Annotation;
  *
  * @Annotation
  * @Target("ANNOTATION")
+ * @Attributes({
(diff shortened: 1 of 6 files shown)


saimaz:
Closing this. We think to integrate variables map directly in mapping collector.

PR #109, opened 2015-01-06
Proxy documents

### LINKED ISSUE #80: generate virtual document classes
Create in cache virtual document classes with setters and getters. If document class already have setter or getter, use it if not create virtual.

martiis:
Closes #80

**DONE**


martiis:
**Clean up stage**. Feel free to review :)

DependencyInjection/ONGRElasticsearchExtension.php
```diff
@@ -39,23 +39,71 @@ public function load(array $configs, ContainerBuilder $container)
...
+     */
+    private function addDocumentFinderDefinition(array $config, ContainerBuilder $container)
+    {
+        $documentFinder = new Definition(
+            'ONGR\ElasticsearchBundle\Mapping\DocumentFinder',
+            [
+                $container->getParameter('kernel.bundles'),
+            ]
+        );
+        $documentFinder->addMethodCall('setDocumentDir', [$config['document_dir']]);
```
saimaz: Since it's most likely last big change in the es bundle, its good time to change `document_dir` to something more meaningful? I mean using shortcuts isn't a good example.

Mapping/Proxy/ProxyLoader.php
```diff
@@ -0,0 +1,107 @@
...
+     * @param \ReflectionClass $reflectionClass
+     *
+     * @return string Proxy document path.
+     */
+    public function load(\ReflectionClass $reflectionClass)
+    {
+        $cacheBundleDir = $this->getCacheDir($reflectionClass->getName());
+
+        if (!$this->getFilesystem()->exists($cacheBundleDir)) {
+            $this->getFilesystem()->mkdir($cacheBundleDir, 0700);
```
saimaz: 0700 means the owner can read, write and execute. Can be that a security vulnerability? @valdas406 what do you think?

valdas406: everything is fine here, directories should have x flag set, but there are some rare cases where permissions should be more strict

martiis:
Should be good

PR #123, opened 2015-01-14
Added initialized flag to document proxies

### LINKED ISSUE #122: Determine if document is on index

martiis:
Closes #122


Mapping/Proxy/ProxyFactory.php
```diff
@@ -101,7 +161,10 @@ private static function getHeader(array $options)
 
 use $base as Base;
 
-class $class extends Base
+/**
+ * DO NOT EDIT THIS FILE - IT WAS CREATED BY ELASTICSEARCH PROXY FACTORY.
+ */
+class $class extends Base implements \ONGR\ElasticsearchBundle\Mapping\Proxy\ProxyInterface
```
saimaz: If we are using `Base`, maybe we should check that project document doesn't use `Base` as the class name before proxy class generation.

PR #130, opened 2015-01-14
add custom RepositoryClass in annotation

juliensantos87:
I add the possibility to have a custom RepositoryClass by annotation like this:

    @ES\Document(type="product", repositoryClass="ProductRepository")

### DIFF
--- a/Annotation/Document.php
+++ b/Annotation/Document.php
@@ -32,6 +32,11 @@ final class Document
     /**
      * @var string
      */
+    public $repositoryClass;
+
+    /**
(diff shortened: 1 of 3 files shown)


martiis:
@juliensantos87 Hi, thanks for PR, but to add a custom repository I would create a factory which could make extra validation or even inject some params from service container. 

With your solution repository could be anything :)

juliensantos87:
@martiis Yes, you're right. I will look for this and travis pass. Thanks

PR #154, opened 2015-01-24
updated travis.yml

martiis:
- Added HHVM build (should be interesting).
 - Updated Elasticsearch versions.
 - PHPCS standard set to ONGR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "mikey179/vfsStream": "~1.4",
         "phpunit/phpunit": "~4.1",
         "squizlabs/php_codesniffer": "~1.5",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -20,7 +20,7 @@
         "mikey179/vfsStream": "~1.4",
         "phpunit/phpunit": "~4.1",
         "squizlabs/php_codesniffer": "~1.5",
-        "ongr/ongr-strict-standard": "~1.0.0-beta"
+        "ongr/ongr-strict-standard": "~1"
```
saimaz: Lets stay with a `~1.0` for a while.

PR #155, opened 2015-01-24
Updated commands with better formated messages

martiis:
Commands more based on connections now will take option connection instead of manager.

Before if we had two managers (foo1 and foo2) with same connections, commands `es:index:create --manager=foo1`, `es:index:create --manager=foo2` would do the same thing. Both commands had the same line  `$connection = $manager->getConnection()` and manager wasn't used anymore. So now commands which are more based on connections will take it as an option.


martiis:
Removed one test case from index create command, because of #156

Command/AbstractConnectionAwareCommand.php
```diff
@@ -0,0 +1,71 @@
...
+namespace ONGR\ElasticsearchBundle\Command;
+
+use ONGR\ElasticsearchBundle\Client\Connection;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
+
+/**
+ * Abstract command which helps executing where connections are required.
+ */
+abstract class AbstractConnectionAwareCommand extends AbstractElasticsearchCommand
```
saimaz: Should be `AbstractManagerAwareCommand`

Command/AbstractConnectionAwareCommand.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->addOption(
+                'connection',
```
saimaz: We are using for all index operations managers. Lets stick to managers naming.

saimaz:
I'm not sure that connection naming is correct. Even in background everything is operating with connections, to pass manager as a command option would make more sense since all index operations are done via managers. Its more unified way.

PR #162, opened 2015-01-28
Added flag to Repository::find()

### LINKED ISSUE #161: Helper method findOrNull needed
When `$repository->find('bad_id');` is executed, exception is thrown. Very good, but helper method should exist, that returns `null`, when document is not found.

I propose `findOrNull` name. Another options: 1. add optional argument 2. provide static helper for wrapping this method.

martiis:
Closes #161

### DIFF
--- a/ORM/Repository.php
+++ b/ORM/Repository.php
@@ -11,6 +11,7 @@
 
 namespace ONGR\ElasticsearchBundle\ORM;
 
+use Elasticsearch\Common\Exceptions\Missing404Exception;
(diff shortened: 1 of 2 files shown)


tautrimas:
Merge conflict @martiis

saimaz:
Proposing to close it. I don't think to putting in all the methods flags is a good way to go.. I agree that exception throwing adds more complexity, so maybe return directly `null` would be better. 

Yes the `null` return is BC break, but in the next minor version is quite fine to release it before stable edition.

PR #163, opened 2015-01-29
Fixing naming conventions

chyzas:


Test/ElasticsearchTestCase.php
```diff
@@ -11,239 +11,11 @@
...
 
-use Elasticsearch\Common\Exceptions\ElasticsearchException;
-use ONGR\ElasticsearchBundle\Client\Connection;
-use ONGR\ElasticsearchBundle\ORM\Manager;
-use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
-use Symfony\Component\DependencyInjection\ContainerInterface;
-
 /**
- * Base test which creates unique connection to test with.
+ * Class ElasticsearchTestCase.
```
tautrimas: Comment too general. We need to disallow usage of `Class` word in doc summaries...

saimaz: Yes we need to disallow usage of `Class`. This file will be removed in stable release so I skipped it. 
@chyzas

PR #174, opened 2015-02-04
AbstractSuggesters [including a bug fix]

### LINKED ISSUE #173: Private properties could not be found from parents in mapping

martiis:
Closes #173


saimaz:
Please do not include any non related bug fixes in the future. If in this particular case abstract suggester won't be good enough the bug fix will be suspended from merge.

PR #175, opened 2015-02-04
Added RESULTS_FIRST flag for repository

### LINKED ISSUE #172: get first result element after execute method
Would be great to have ability to get first element of query. First option is to add additional flag to execute f.g. `RESULT_ARRAY_FIRST`

However maybe its better to integrate additional method in the `Iterator` class.

```
$firstResult = $repository->execute($search)->first();
```

martiis:
Closes #172


saimaz:
To add first result as a separate response type I'm not sure is the best way to solve the issue. Much better approach would be to add first method to the iterator. 

If someone does not agree we can reopen issue and discuss about it.

PR #196, opened 2015-02-23
mappings are now able to accept separate documents

### LINKED ISSUE #185: Mappings should be able to accept separate documents
Ability to add only selected documents to specified index is needed, currently, all documents from one bundle are added to the same index. Sometimes, we need that documents have to be on separate indexes.

```
ongr_elasticsearch:
    ...
    managers:
        default:
            ...
            mappings:
                - ONGRMatchMergeBundle:Settings
```

martiis:
Closes #185


martiis:
To keep it tight and clean I had to change one method name:
  - *MetadataCollector#getMapping* -> *MetadataCollector#getClientMapping*

Now with *MetadataCollector#getMapping* method bundle or document mapping with metadata is retrieved.

PR #200, opened 2015-02-24
Added Manager option to be readonly.

### LINKED ISSUE #149: Consider implementing readonly connections/managers
In some cases it would be useful to have readonly managers.

Would be nice to have option on connection and/or on manager to set it to readonly.

linasmo:
Closes #149


martiis:
I would more likely set it through setter not constructor.

linasmo:
Setting moved to setter. 
Readonly checks updated.

PR #204, opened 2015-02-25
Refactored code

chyzas:


DSL/Query/ConstantScoreQuery.php
```diff
@@ -29,16 +32,17 @@ class ConstantScoreQuery implements BuilderInterface
...
+    private $query;
 
     /**
-     * @param BuilderInterface $filterOrQuery
+     * @param BuilderInterface $query
      * @param array            $parameters
+     * @param string           $dslType
      */
-    public function __construct(BuilderInterface $filterOrQuery, array $parameters = [])
+    public function __construct(BuilderInterface $query, array $parameters = [], $dslType = self::USE_FILTER)
```
linasmo: Maybe default value should be ``self::USE_QUERY`` since it is ``ConstantScoreQuery`` query.

DSL/Query/FunctionScoreQuery.php
```diff
@@ -37,14 +40,19 @@ class FunctionScoreQuery implements BuilderInterface
...
      */
-    public function __construct(BuilderInterface $filterOrQuery, array $functions, array $parameters = [])
-    {
-        $this->dslType = array_slice(explode('\\', get_class($filterOrQuery)), -2, 1)[0];
-        $this->filterOrQuery = $filterOrQuery;
+    public function __construct(
+        BuilderInterface $query,
+        array $functions,
+        array $parameters = [],
+        $dslType = self::USE_FILTER
```
linasmo: Maybe default value should be ``self::USE_QUERY`` since it is ``FunctionScoreQuery `` query.

DSL/Query/MatchQuery.php
```diff
@@ -24,22 +24,22 @@ class MatchQuery implements BuilderInterface
...
+    private $query;
 
     /**
-     * @param string $query
      * @param string $field
+     * @param string $query
      * @param array  $parameters
      */
-    public function __construct($query, $field, array $parameters = [])
+    public function __construct($field, $query, array $parameters = [])
```
linasmo: Why parameters was switched places?

DSL/Query/MultiMatchQuery.php
```diff
@@ -19,9 +19,14 @@
 class MultiMatchQuery implements BuilderInterface
 {
     /**
-     * @var string
+     * @param array  $fields
+     * @param string $query
      */
-    private $query;
+    public function __construct(array $fields, $query)
```
linasmo: Why constructor was moved above private properties?

martiis:
I would suggest to put constants into one place because they are all the same.

DSL/Query/MatchQuery.php
saimaz: Its for unifying reasons. All other filters and queries uses field and value order.

saimaz: However match query doesn't have field property. @chyzas please remove it.

PR #215, opened 2015-02-27
Added missing filters

chyzas:
geo distance, geo distance range, geo polygon filters

### DIFF
--- a/DSL/Filter/GeoDistanceFilter.php
+++ b/DSL/Filter/GeoDistanceFilter.php
@@ -24,12 +24,12 @@ class GeoDistanceFilter implements BuilderInterface
     /**
      * @var string
      */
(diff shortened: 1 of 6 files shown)


DSL/Filter/GeoDistanceFilter.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    private $location;
+
+    /**
+     * @param string $distance
+     * @param string $field
+     * @param mixed  $location
+     * @param array  $parameters
+     */
+    public function __construct($distance, $field, $location, array $parameters = [])
```
saimaz: First shoud be `$field`, then distance.

saimaz:
Docs are missing.

PR #230, opened 2015-03-04
Tests for Completion setters/getters and Completion::toArray exception

### LINKED ISSUE #225: Test missing for DSL\Suggester\Completion
https://github.com/ongr-io/ElasticsearchBundle/blob/master/DSL/Suggester/Completion.php

AndriusBil:
Closes #225

### DIFF
--- a/Test/EncapsulationTestAwareTrait.php
+++ b/Test/EncapsulationTestAwareTrait.php
@@ -119,7 +119,7 @@ trait EncapsulationTestAwareTrait
 
         $stub = $this->getStub();
 
-        if (!isset($stub)) {
+        if ($stub === null) {
             $stub = $this->getMockForAbstractClass($objectClass);
         }
 
--- a/Test/EncapsulationTestAwareTrait.php
+++ b/Test/EncapsulationTestAwareTrait.php
@@ -9,14 +9,14 @@
  * file that was distributed with this source code.
  */
 
-namespace ONGR\ElasticsearchBundle\Tests\app\fixture;
-
-use PHPUnit_Framework_MockObject_MockObject;
+namespace ONGR\ElasticsearchBundle\Test;
 
 /**
- * Abstract entity test for setters and getters.
+ * A helper for testing setters/getters.
+ *
+ * Trait EncapsulationTestAwareTrait
  */
-abstract class AbstractEntityTest extends \PHPUnit_Framework_TestCase
+trait EncapsulationTestAwareTrait
 {
     /**
      * @var string[] List of fields that should not be checked for tests.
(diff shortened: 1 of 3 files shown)


Tests/app/fixture/AbstractEntityTest.php
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * (c) NFQ Technologies UAB <info@nfq.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ONGR\ElasticsearchBundle\Tests\app\fixture;
+
+use PHPUnit_Framework_MockObject_MockObject;
```
saimaz: No need to define use case like this.

Tests/app/fixture/AbstractEntityTest.php
```diff
@@ -0,0 +1,269 @@
...
+/*
+ * This file is part of the ONGR package.
+ *
+ * (c) NFQ Technologies UAB <info@nfq.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ONGR\ElasticsearchBundle\Tests\app\fixture;
```
saimaz: Are you sure it is the right namespace and place for such a kind `abstract` test? Do we need this test class at all?

Test/EncapsulationTestAwareTrait.php
```diff
@@ -0,0 +1,277 @@
...
+        $setter = 'set' . ucfirst($field);
+        $getter = 'get' . ucfirst($field);
+
+        if ($type === 'boolean') {
+            $getter = 'is' . ucfirst($field);
+        }
+
+        $stub = $this->getStub();
+
+        if (!isset($stub)) {
```
tautrimas: This seem to be always `false`.

AndriusBil: isset() will return FALSE if testing a variable that has been set to NULL (php.net manual).
$this->stub is set to null by default. In case we don't change $this->stub value !isset($this->stub) should generate true. Or am I getting something wrong ?

tautrimas: OK, then simply do `$stub === null` check.

Test/EncapsulationTestAwareTrait.php
```diff
@@ -0,0 +1,277 @@
...
+ */
+
+namespace ONGR\ElasticsearchBundle\Test;
+
+/**
+ * A helper for testing setters/getters.
+ *
+ * Trait EncapsulationTestAwareTrait
+ */
+trait EncapsulationTestAwareTrait
```
tfe2012: why you use trait instead of abstract class?

AndriusBil: There might be scenarios, where we need to test encapsulation, but we're already using abstraction class.

PR #237, opened 2015-03-05
Added filters, geo bound, geohash grid aggregations

chyzas:


DSL/Aggregation/GeoHashGridAggregation.php
```diff
@@ -0,0 +1,121 @@
...
+    {
+        $data = [];
+
+        if ($this->getField()) {
+            $data['field'] = $this->getField();
+        } else {
+            throw new \LogicException('Geo bounds aggregation must have a field set.');
+        }
+
+        if ($this->precision) {
```
martiis: If you have getters, always use them.

PR #240, opened 2015-03-06
Added span first, span near, span no, spar or queries

chyzas:


DSL/Query/SpanNearQuery.php
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @param array $queries
+     * @param array $parameters
+     *
+     * @throws \LogicException
+     */
+    public function __construct(array $queries = [], array $parameters = [])
+    {
+        foreach ($queries as $query) {
```
martiis: Couldn't just `$this->queries = $queries`?

DSL/Query/SpanNearQuery.php
```diff
@@ -0,0 +1,69 @@
...
+{
+    use ParametersTrait;
+
+    /**
+     * @var BuilderInterface[]
+     */
+    private $queries = [];
+
+    /**
+     * @param array $queries
```
martiis: not an `array` but `BuilderInterface[]`

DSL/Query/SpanFirstQuery.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param BuilderInterface $query
+     * @param array            $parameters
+     *
+     * @throws \LogicException
+     */
+    public function __construct($query, array $parameters = [])
+    {
+        $this->query = $query;
+        $this->setParameters($parameters);
+        if (!$this->hasParameter('end')) {
```
saimaz: `end` is mandatory field so it must be in the constructor.

DSL/Query/SpanFirstQuery.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    private $query;
+
+    /**
+     * @param BuilderInterface $query
+     * @param array            $parameters
+     *
+     * @throws \LogicException
+     */
+    public function __construct($query, array $parameters = [])
```
saimaz: `$query` must have `BuilderInterface` type hint

PR #252, opened 2015-03-10
Refactored query building

saimaz:
From now on bool operator will be added only if there is more than one query, or other conditions than must. This PR is BC compatible, no Tests changed, except not necessary bool removal.


saimaz:
BTW with the same behavior we could improve filter usage.

saimaz:
@mindaugasbarysas @tautrimas can you check what is wrong with sphinx here?

mindaugasbarysas:
"default" theme has been renamed to "classic". I've updated config accordingly.

PR #261, opened 2015-03-13
Unit tests for AbstractDocument class

AndriusBil:
Unit tests for Document/AbstractDocument


saimaz:
Well with this test case you had to find that code isn't correct. Actually I've missed to add abstract class type in `AbstractDocument` class. Normally you can't create abstract class objects, for that have to use mock for abstract class.

Please fix also `AbstractClass` by adding abstract type.

PR #262, opened 2015-03-13
Unit tests for converter class

AndriusBil:
Unit tests for Result/Converter

### DIFF
--- a/Tests/Unit/Result/ConverterTest.php
+++ b/Tests/Unit/Result/ConverterTest.php
@@ -39,7 +39,7 @@ class ConverterTest extends \PHPUnit_Framework_TestCase
      */
     public function testAssignArrayToObject()
     {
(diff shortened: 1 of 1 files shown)


Tests/Unit/Result/ConverterTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $converter = new Converter([], []);
+        $converter->convertToDocument(['_type' => 'foo']);
+    }
+
+    /**
+     * Tests assignArrayToObject().
+     */
+    public function testAssignArrayToObject()
+    {
+        $stub = new \StdClass();
```
saimaz: This should be at least document class mock with setter stubs. We want to test if converter actually calls setters.

saimaz:
This article should be useful to read for future: http://stopwritingramblingcommitmessages.com

PR #263, opened 2015-03-13
Additional tests for ElasticsearchExtension class

AndriusBil:
Tests for DependencyInjection/ElasticsearchExtension class


Tests/Unit/DependencyInjection/ElasticsearchExtensionTest.php
```diff
@@ -65,6 +94,75 @@ public function testLoad()
...
+                    'template' => 'ONGRElasticsearchBundle:Profiler:profiler.html.twig',
+                    'id' => 'es',
+                ]
+            );
+
+            $this->assertEquals(
+                $collector,
+                $container->getDefinition('es.collector')
+            );
+            $this->assertEquals(
```
saimaz: More accurate would be to place assert after definition. But for now its ok.

PR #264, opened 2015-03-13
Additional tests for IndexCreateCommand

AndriusBil:


Tests/Functional/Command/CreateIndexCommandTest.php
```diff
@@ -35,7 +42,7 @@ public function getTestExecuteData()
             [
                 'default',
                 [
-                    'timestamp' => false,
+                    'timestamp' => true,
```
saimaz: It doesn't look like additional test when you change current data provider. Please add as real additional statement.

PR #265, opened 2015-03-16
Added geo-shape filter

chyzas:
This filter is divided in two filters: GeoShape provided and GeoShape pre-indexed.


saimaz:
All filters must have `Filter` suffix. Next pre-indexed filter variant is the same geo_shape filter, implement it in the `GeoShapeFilter`. Same with `GeoShape` provided.

PR #267, opened 2015-03-16
Removed unnecessary parameters building in dev env.

saimaz:


Tests/app/config/parameters_test.yml
```diff
@@ -0,0 +1,5 @@
+# This file is auto-generated during the composer install
```
martiis: This comment should not exist?

PR #269, opened 2015-03-16
DSL Span queries

chyzas:


DSL/Query/Span/SpanNearQuery.php
```diff
@@ -0,0 +1,70 @@
...
+    private $queries = [];
+
+    /**
+     * @param int                  $slop
+     * @param SpanQueryInterface[] $queries
+     * @param array                $parameters
+     *
+     * @throws \LogicException
+     */
+    public function __construct($slop, array $queries = [], array $parameters = [])
```
saimaz: Remove `$queries` here. Queries should be added via `addQuery(SpanQueryInterface $query)`.

DSL/Query/Span/SpanNearQuery.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function toArray()
+    {
+        $query = [];
+        foreach ($this->queries as $type) {
+            $data = [$type->getType() => $type->toArray()];
```
saimaz: `$data` is abstract keyword which should not be used. Change it to `$queries`. Check also in other places same cases.

DSL/Query/Span/SpanOrQuery.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @var SpanQueryInterface[]
+     */
+    private $queries = [];
+
+    /**
+     * @param SpanQueryInterface[] $queries
+     * @param array                $parameters
+     */
+    public function __construct(array $queries = [], array $parameters = [])
```
saimaz: Same case as in `SpanNearQuery` with `$queries`.

PR #280, opened 2015-03-19
Patch dsl unit tests

mansimas:
Added aggregation tests.

### DIFF
--- a/Tests/Unit/DSL/Aggregation/GeoBoundsAggregationTest.php
+++ b/Tests/Unit/DSL/Aggregation/GeoBoundsAggregationTest.php
@@ -46,14 +46,26 @@ class GeoBoundsAggregationTest extends \PHPUnit_Framework_TestCase
     {
         $agg = new GeoBoundsAggregation('foo');
         $agg->setField('bar');
-        $agg->setWrapLongitude(false);
-        $result = $agg->getArray();
-        $this->assertArrayHasKey('field', $result);
-        $this->assertArrayHasKey('wrap_longitude', $result);
-        $this->assertEquals('bar', $result['field']);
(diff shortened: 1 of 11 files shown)


Tests/Unit/DSL/Aggregation/GeoBoundsAggregationTest.php
```diff
@@ -28,4 +28,32 @@ public function testGeoBoundsAggregationException()
...
+    public function testGeoBoundsAggregationGetArray()
+    {
+        $agg = new GeoBoundsAggregation('foo');
+        $agg->setField('bar');
+        $agg->setWrapLongitude(false);
+        $result = $agg->getArray();
+        $this->assertArrayHasKey('field', $result);
+        $this->assertArrayHasKey('wrap_longitude', $result);
+        $this->assertEquals('bar', $result['field']);
+        $this->assertFalse($result['wrap_longitude']);
```
martiis: When test has more than one assertion try to enter a third parameter (message) so that it would be easier to know later which assertion failed

Tests/Unit/DSL/Aggregation/RangeAggregationTest.php
```diff
@@ -118,24 +120,93 @@ public function getToArrayData()
...
+        $result = $agg->getType();
+        $this->assertEquals('range', $result);
+    }
+
+    /**
+     * Data provider for testRangeAggregationRemoveRangeByKey(), testRangeAggregationRemoveRange().
+     *
+     * @return array
+     */
+    public function testRangeAggregationDataProvider()
```
martiis: Data providers should not only give expected results, but also data which you would give to objects/process and then check if the expected data matches with returned from object/process.

PR #282, opened 2015-03-19
Renamed FuzzyLikeThisFieldQuery

chyzas:


DSL/Query/DisMaxQuery.php
```diff
@@ -29,24 +29,28 @@ class DisMaxQuery implements BuilderInterface
...
-        }
         $this->setParameters($parameters);
     }
 
     /**
+     * Add query.
+     *
+     * @param BuilderInterface $query
+     *
+     * @return $this
```
martiis: It returns `DisMaxQuery`

PR #288, opened 2015-03-23
Fixes IndexPrimaryShardNotAllocatedException

chyzas:


martiis:
This fix is not correct

PR #292, opened 2015-03-24
Added DSL unit tests.

dvondrak:


Tests/Unit/DSL/SearchTest.php
```diff
@@ -0,0 +1,169 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ONGR\ElasticsearchBundle\Tests\Unit\DSL;
+
+use ONGR\ElasticsearchBundle\DSL\Search;
+use ONGR\ElasticsearchBundle\DSL\Sort\AbstractSort;
+
+class SearchTest extends \PHPUnit_Framework_TestCase
```
martiis: docBlock is missing. *check others*

Tests/Unit/DSL/SearchTest.php
```diff
@@ -0,0 +1,169 @@
...
+        return $mock;
+    }
+
+    /**
+     * Tests addSort method.
+     */
+    public function testAddSort()
+    {
+        $search = new Search();
+        $this->assertEmpty($search->getSorts());
```
martiis: When test has more than one assertion try to enter a third parameter (message) so that it would be easier to know later which assertion failed. **check other places also**

PR #295, opened 2015-03-24
Tests

LTRocky:


DSL/Search.php
```diff
@@ -69,7 +69,7 @@ class Search
     /**
      * @var string|null
      */
-    private $scrollDuration;
+    private $scroll;
```
martiis: `$scrollDuration` was correct naming IMO. Why did you change it?

DSL/Search.php
```diff
@@ -223,6 +223,62 @@ public function setFields(array $fields)
     }
 
     /**
+     * Set filters.
+     *
+     * @param BuilderInterface[] $filters
+     *
+     * @return $this
+     */
+    public function setFilters($filters)
```
martiis: I don't think this is correct. Because we contain filters in bool container and it is handled by search object. This setter is also incorrect because it says that parameter `$filters` is a `BuilderInterface[]` but it should be `Bool`. Same goes to setter for post filters. For aggregations and suggesters it's a no also. We intentionally wrapped them in Search object because they use same container type and by not exposing them could have some changes internally, which user would not even notice.

DSL/Search.php
LTRocky: First of all $scrollDuration is a private field. It has a getter getScroll() and a setter setScroll(). So is it really a bad thing to rename it to simply $scroll ? I've changed its name in order to be able to test its getters and setters using EncapsulationTestAwareTrait.

martiis: Okay.

DSL/Search.php
martiis: Please remove these setters, because they are keeping me against merging.

DSL/Search.php
saimaz: `scroll` is completely fine. Even in elasticsearch itself this parameter is named `scroll` not `scrollDuration`

Tests/Unit/DSL/Filter/PostFilterTest.php
```diff
@@ -32,4 +32,40 @@ public function testIfIsRelevantFunctionIsReturningFalse()
         $bool = new PostFilter();
         $this->assertFalse($bool->isRelevant());
     }
+
+    /**
+     * Test addFilter method.
+     */
+    public function testAddFilter()
+    {
+        $filterMock = $this->getMockBuilder('ONGR\ElasticsearchBundle\DSL\Filter\MissingFilter')
```
saimaz: `$filterMock` must be named `$missingFilterMock`

Tests/Unit/DSL/Filter/PostFilterTest.php
```diff
@@ -32,4 +32,40 @@ public function testIfIsRelevantFunctionIsReturningFalse()
...
+            ->setMethods(['addToBool'])
+            ->disableOriginalConstructor()
+            ->getMock();
+        $filterMock->expects($this->once())
+            ->method('addToBool')
+            ->withAnyParameters();
+
+        $filter = new PostFilter();
```
saimaz: `$filter` must be named `$postFilter`

Tests/Unit/DSL/Filter/PostFilterTest.php
```diff
@@ -32,4 +32,40 @@ public function testIfIsRelevantFunctionIsReturningFalse()
...
+        $filter->setFilter($filterMock);
+        $filter->addFilter($filterMock, 'test');
+    }
+
+    /**
+     * Test setBoolParameters method.
+     */
+    public function testSetBoolParameters()
+    {
+        $filterMock = $this->getMockBuilder('ONGR\ElasticsearchBundle\DSL\Filter\MissingFilter')
```
saimaz: `$filterMock` must be named `$missingFilterMock `

Tests/Unit/DSL/Filter/PostFilterTest.php
```diff
@@ -32,4 +32,40 @@ public function testIfIsRelevantFunctionIsReturningFalse()
...
+            ->setMethods(['setParameters'])
+            ->disableOriginalConstructor()
+            ->getMock();
+        $filterMock->expects($this->once())
+            ->method('setParameters')
+            ->withAnyParameters();
+
+        $filter = new PostFilter();
```
saimaz: `$filter` must be named `$postFilter`

Tests/Unit/DSL/ParametersTraitTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Test for ParametersTrait.
+ */
+class ParametersTraitTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var ParametersTrait
+     */
+    private $mock;
```
saimaz: Name is incorrect. Use more defining names. `parametersTraitMock`should be better.

PR #296, opened 2015-03-24
Added Unit/DSL/Query/Span tests

mansimas:


Tests/Unit/DSL/Query/Span/SpanFirstQueryTest.php
```diff
@@ -0,0 +1,77 @@
...
+            ->method('toArray')
+            ->will($this->returnValue(['key' => 'value']));
+    }
+
+    /**
+     * Reset mock object.
+     */
+    public function tearDown()
+    {
+        $this->mock = null;
```
martiis: If you want to save memory use `unset()` because `= null` does not remove variable from symbol table.

PR #298, opened 2015-03-24
Change ONGR to Ongr in namespaces and files

### LINKED ISSUE #239: Change ONGR to Ongr in namespaces and files
Also please take a look if there is another camel case issues. Keep in mind that settings keys must be lower cased.

LTRocky:
Closes #239


saimaz:
Currently closing to avoid accidentally merge. Will merge later when it will be finally planned.

PR #308, opened 2015-04-01
Code coverage should only be executed once.

Zylius:
Turns out we will need to run PHPUnit with xDebug at least once to get code coverage. For example, to get code coverage from Scrutinizer you must use its CI which will also run PHPUnit multiple times, CoverAlls require you to send data to it from CI like we're sending it now to Scrutinizer, wasn't able to find a tool which works differently.

So I took an idea from Doctrine https://github.com/doctrine/doctrine2/commit/934e4fdde4e07d464a0041117732e908dde3cbde, but instead of enabling code coverage on a specific version, I disabled xDebug (which also disables code coverage) on all except one version (which is specified by ES version and PHP version).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ install:
   - wget -q -O conf.py https://raw.githubusercontent.com/ongr-io/docs-aggregator/master/source/conf-travis.py
 
 before_script:
+  - if [[ $TRAVIS_PHP_VERSION = 5.6 && $ELASTICSEARCH = 1.4.4 ]]; then COVERAGE_NEEDED=true; else COVERAGE_NEEDED=false; fi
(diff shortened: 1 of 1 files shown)


saimaz:
Something is wrong. Test time amount is still the same (more or less). With disabled xdebug it should cut time atleast twice.

Zylius:
But xDebug is disabled. You get `The Xdebug extension is not loaded. No code coverage will be generated.` message in all builds except [this one](https://travis-ci.org/ongr-io/ElasticsearchBundle/jobs/56726748).

saimaz:
Ok, I'll try to restart travis builds. Lets see if it will make difference. Rigth now it takes 1 hr 11 min 13 sec.

Zylius:
Waiting for workers takes most of that time (especially PHP7 and HHVM), there should be a clear difference between xDebug build and non-xDebug builds nonetheless.

PR #320, opened 2015-04-07
Added histogram aggregation

chyzas:


DSL/Aggregation/HistogramAggregation.php
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * Class representing Histogram aggregation.
+ */
+class HistogramAggregation extends AbstractAggregation
+{
+    use BucketingTrait;
+
+    const DIRECTION_ASC = 'asc';
+    const DIRECTION_DESC = 'desc';
```
martiis: `DIRECTION_DESC` is a bit misleading, `SORT_DESC` would be more direct.

PR #325, opened 2015-04-08
Updated terms aggregation

chyzas:


DSL/Aggregation/TermsAggregation.php
```diff
@@ -19,11 +20,16 @@
 class TermsAggregation extends AbstractAggregation
 {
     use BucketingTrait;
+    use ScriptAwareTrait;
 
-    const MODE_COUNT = '_count';
-    const MODE_TERM = '_term';
```
martiis: have they been taken out from es docs?

chyzas: No. But you can use your own order mode too, for example
```json
{
    "aggs" : {
        "genders" : {
            "terms" : {
                "field" : "gender",
                "order" : { "avg_height" : "desc" }
            },
            "aggs" : {
                "avg_height" : { "avg" : { "field" : "height" } }
            }
        }
    }
}
```

Issue #220, opened 2015-03-02
field value evaluator, to get nested values from document


trylika:
There is need for functionality, to get any nested value from `Document`, by passing string depicting value requested. I.e.

`personDocument.getValue('work.address.street')`

would get from Person Document, it's work object's street name.

martiis:
Try using http://symfony.com/doc/current/components/property_access/introduction.html#reading-from-arrays

saimaz:
Would be nice inject `PropertyAccess` in the proxy class.

saimaz:
I think that `getValue` method is not the best choice, because there is sligthly high chance that user will create a document with `$value` property, and then we will override his getter...

lmikelionis:
I would call that method: getValueFromField()

PR #327, opened 2015-04-08
Impletemented functionality to get any nested value from Document

chyzas:


chyzas:
closes #220

saimaz:
Write additional test for non multiple document property.

PR #333, opened 2015-04-09
Patch scrutinizer

chyzas:


DSL/Aggregation/TopHitsAggregation.php
```diff
@@ -49,7 +49,9 @@ public function __construct($name, $size = null, $from = null, $sort = null)
         parent::__construct($name);
         $this->setFrom($from);
         $this->setSize($size);
-        $this->setSort($sort);
+        if (!empty($sort)) {
```
martiis: This fix should not be in constructor, but in toArray method.

PR #338, opened 2015-04-13
added simple installation guide

mansimas:


Resources/doc/guide.rst
```diff
@@ -0,0 +1,211 @@
...
+        managers:
+            default:
+                connection: default
+                mappings:
+                    - AcmeDemoBundle
+
+Creating Database for ElasticSearch
+-----------------------------------
+
+Go to ``src/Acme/DemoBundle`` and create new Folder ``Document``. It must be named as "Document". ONGR points to exactly this directory.
```
linasmo: Its not entirely true. Document directory can be specified in config.
https://github.com/ongr-io/ElasticsearchBundle/blob/master/DependencyInjection/Configuration.php#L33

mansimas: I plan to include this later, how to change Document's folder.  I could edit this line like "ONGR points to exactly this directory as default".

Resources/doc/guide.rst
```diff
@@ -0,0 +1,211 @@
...
+.. code:: bash
+
+    mkdir Sites
+    cd Sites
+
+Create new Symfony project
+
+.. code:: bash
+
+    symfony new project
```
martiis: This `symfony` command is available by using https://github.com/symfony/symfony-installer. It is recommended way but also mention composer `composer create-project symfony/framework-standard-edition`

martiis:
Else looks nice :+1:

PR #350, opened 2015-04-20
Scan iterator is now clearing memory by itself

### LINKED ISSUE #336: Memory leak DocumentScanIterator
Then you scan in offsetGet($offset) 

https://github.com/ongr-io/ElasticsearchBundle/blob/master/Result/AbstractResultsIterator.php#L89

grow $this->documents and $this->converted

martiis:
From https://github.com/ongr-io/ElasticsearchBundle/tree/remove_memory_leak

Closes #336

### DIFF
--- a/Result/DocumentScanIterator.php
+++ b/Result/DocumentScanIterator.php
@@ -130,4 +130,29 @@ class DocumentScanIterator extends DocumentIterator
     {
         $this->key++;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function offsetGet($offset)
+    {
+        if (!isset($this->converted[$offset])) {
+            if (!isset($this->documents[$offset])) {
+                return null;
+            }
+
+            $this->converted[$offset] = $this->convertDocument($this->documents[$offset]);
+
+            // Clear memory.
+            if (isset($this->converted[$offset - 20])) {
+                $this->converted[$offset - 20] = null;
+            }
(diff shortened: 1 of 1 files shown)


Result/DocumentScanIterator.php
```diff
@@ -130,4 +130,29 @@ public function next()
...
+    {
+        if (!isset($this->converted[$offset])) {
+            if (!isset($this->documents[$offset])) {
+                return null;
+            }
+
+            $this->converted[$offset] = $this->convertDocument($this->documents[$offset]);
+
+            // Clear memory.
+            if (isset($this->converted[$offset - 20])) {
```
saimaz: For what reason is `20`.

saimaz:
Moving to `1.0.0` milestone. From first look I dont see any memory leaks here. We will investigate this additionally.

Result/DocumentScanIterator.php
tautrimas: @tfe2012 should know the exact reason, but I believe this was a hack. Integer `1` or even `2` should be enough, because you cannot iterate the scan iterator again anyways.

tfe2012: it's random number.

saimaz:
We are leaving issue open but clodding PR, and will try to reduce memory leak as much as possible in `1.0`.

PR #355, opened 2015-04-21
Fixes phpcs2 concatenate issues

chyzas:


martiis:
This case needs to be corrected in strict standard

PR #366, opened 2015-04-24
added custom LogicException to Connections commit method if nothing pers...

### LINKED ISSUE #356: Bad request exception on commit before persisting anything
If we try to commit before persisting anything (as we have situation when we do not always know that we persisted anything), we get bad request exception:
 ```
ElasticsearchParseException[Failed to derive xcontent from org.elasticsearch.common.bytes.BytesArray@1]
```

Error occurs when the following line returns empty result and we try to call bulk query with empty body:
https://github.com/ongr-io/ElasticsearchBundle/blob/9b18076e1af8f47fa61cc5e659effc6b3630ced4/Client/Connection.php#L156

Consider adding result check before calling bulk query or maybe throw custom exception?

### PR DESCRIPTION
**Pull request by mansimas**
...isted

Closes #356

### DIFF
--- a/Client/Connection.php
+++ b/Client/Connection.php
@@ -16,7 +16,7 @@ use Elasticsearch\Common\Exceptions\Forbidden403Exception;
 use ONGR\ElasticsearchBundle\Cache\WarmerInterface;
 use ONGR\ElasticsearchBundle\Cache\WarmersContainer;
 use ONGR\ElasticsearchBundle\Mapping\MappingTool;
-use ONGR\ElasticsearchBundle\Exceptions\EmptyBulkException;
+use ONGR\ElasticsearchBundle\Exception\EmptyBulkException;
 
 /**
  * This class interacts with elasticsearch using injected client.
@@ -157,7 +157,7 @@ class Connection
     public function commit()
     {
         if (!$this->bulkQueries) {
-            throw new EmptyBulkException('Nothing persisted to commit.');
+            throw new EmptyBulkException('Bulk container is empty. Add some queries before commit.');
         }
         $this->bulkQueries = array_merge($this->bulkQueries, $this->bulkParams);
         $this->getClient()->bulk($this->bulkQueries);
--- a/Client/Connection.php
+++ b/Client/Connection.php
@@ -150,13 +150,17 @@ class Connection
 
     /**
      * Flushes the current query container to the index, used for bulk queries execution.
+     *
+     * @throws \LogicException
      */
     public function commit()
     {
+        if (!$this->bulkQueries) {
+            throw new \LogicException('Nothing persisted to commit.');
+        }
         $this->bulkQueries = array_merge($this->bulkQueries, $this->bulkParams);
         $this->getClient()->bulk($this->bulkQueries);
         $this->flush();
-
         $this->bulkQueries = [];
     }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Exceptions/EmptyBulkException.php [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+/*
+ * This file is part of the ONGR package.
+ *
+ * (c) NFQ Technologies UAB <info@nfq.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace ONGR\ElasticsearchBundle\Exceptions;
```
saimaz: Right name space would be `namespace ONGR\ElasticsearchBundle\Exception`

Client/Connection.php
```diff
@@ -150,13 +151,17 @@ public function setBulkParams(array $params)
 
     /**
      * Flushes the current query container to the index, used for bulk queries execution.
+     *
+     * @throws EmptyBulkException
      */
     public function commit()
     {
+        if (!$this->bulkQueries) {
+            throw new EmptyBulkException('Nothing persisted to commit.');
```
saimaz: Right message would be: `Bulk container is empty. Add some queries before commit.`

Tests/Unit/Client/ConnectionTest.php
```diff
@@ -156,10 +156,34 @@ public function testGetClient()
...
+        $connection = new Connection($client, ['index' => 'foo']);
         $connection->setBulkParams(['refresh' => 'true']);
+        $connection->bulk('create', 'type', []);
+        $connection->commit();
+    }
+
+    /**
+     * Tests if exception is thown when only commit is made.
+     *
+     * @expectedException \LogicException
```
saimaz: Should expect `EmptyBulkException`

Tests/Unit/Client/ConnectionTest.php
```diff
@@ -156,10 +156,34 @@ public function testGetClient()
...
 
-        $connection = new Connection($client, []);
+        $connection = new Connection($client, ['index' => 'foo']);
         $connection->setBulkParams(['refresh' => 'true']);
+        $connection->bulk('create', 'type', []);
+        $connection->commit();
+    }
+
+    /**
+     * Tests if exception is thown when only commit is made.
```
saimaz: Check spelling.

Tests/Unit/Client/ConnectionTest.php
```diff
@@ -156,10 +156,34 @@ public function testGetClient()
...
+                0 => [
+                    'create' => [
+                        '_index' => 'foo',
+                        '_type' => 'type',
+                    ],
+                ],
+                1 => [],
+            ],
+        ];
+        $client->expects($this->once())->method('bulk')->with($result);
```
saimaz: This is nothing related to empty bulk exception functionality. Please do it as a separate PR. Remove these lines with rebase to keep history clean.

saimaz:
@mansimas any idea when you have time to fix your PR?

mansimas:
I can do it today.

Tests/Unit/Client/ConnectionTest.php
mansimas: I cant revert this, because after my changes it fails.

PR #371, opened 2015-04-28
added format property to Property annotation

martiis:


tautrimas:
Is it possible, that format for Elasticsearch is not the same as for PHP / C languages?

Also, will this PR pass the `format` parameter to Elasticsearch type mapping?

martiis:
This will pass format to mapping. For formats check this http://joda-time.sourceforge.net/api-release/org/joda/time/format/DateTimeFormat.html

saimaz:
Elasticsearch uses org.joda.time.format from Java.

tautrimas:
Yes, they are incompatible. ES will use 'y' for 4 digit year, and PHP will want 'Y' for 4 digit year. Maybe do not use annotation format for parsing the date on PHP side? PHP parser is pretty intelligent anyways.

PR #381, opened 2015-05-13
Manager: typehints added

TomasVotruba:


martiis:
Hi @TomasVotruba, thanks for the touches, but we had unit some tests that did not use Connection and just `null` was passed. Would be great if they would be changed also!

TomasVotruba:
@martiis Sure. I'll look on it.

saimaz:
As @martiis wrote this is the reason and not the only one why we didn't add type hinting. Closing for now. @TomasVotruba If you will find a better solution feel free to reopen.

TomasVotruba:
@saimaz For there use cases, null objects are used.

PR #383, opened 2015-05-15
Repository service uses factory method

tautelis:
Situation:
```php
$manager = $client->getContainer()->get('es.manager.default');
$manager->getConnection()->getIndexName();
// Returns 'index'
$manager->getConnection()->setIndexName('new-index');
$manager->getConnection()->getIndexName();
// Returns 'new-index'
$repository = $client->getContainer()->get('es.manager.default.category');
$repository->getManager()->getConnection()->getIndexName();
// Returns 'index', tough it should have returned 'new-index'
```
An index name change applied for the manager does not apply to the repository services. This happens because each repository Definition has it's own parameters cached in Container cache. Generating repositories by factory fixes this issue.


saimaz:
Thank you. This was a really good finding.

Issue #346, opened 2015-04-17
Add new method to Repository class


tfe2012:
method find return DocumentInterface

method findBy return DocumentIterator

I think it would be useful create method findOneBy which return DocumentInterface

asev:
@tfe2012 is this what have you expected? Can it be closed now?

tautelis:
This one could be closed for sure, method findOneBy is already there and used a lot :)

PR #385, opened 2015-05-16
findOneBy added to Repository

asev:
This is useful method for repository.

### DIFF
--- a/ORM/Repository.php
+++ b/ORM/Repository.php
@@ -162,12 +162,13 @@ class Repository
     /**
      * Finds only one entity by a set of criteria.
      *
-     * @param array      $criteria Example: ['group' => ['best', 'worst'], 'job' => 'medic'].
-     * @param array|null $orderBy  Example: ['name' => 'ASC', 'surname' => 'DESC'].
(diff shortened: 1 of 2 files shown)


ORM/Repository.php
```diff
@@ -160,6 +160,32 @@ public function findBy(
...
 
     /**
+     * Finds only one entity by a set of criteria.
+     *
+     * @param array      $criteria Example: ['group' => ['best', 'worst'], 'job' => 'medic'].
+     * @param array|null $orderBy  Example: ['name' => 'ASC', 'surname' => 'DESC'].
+     *
+     * @return DocumentInterface|null The object.
+     */
+    public function findOneBy(array $criteria, array $orderBy = [])
```
saimaz: IMO there should be a third argument with result type.

martiis:
Issue #346

tautelis:
Looking forward for the release with this useful method!

asev:
I've added resultType argument to findOneBy(). In case when resultType is RESULTS_OBJECT - it acts the same as find().

asev:
This one is still important for our project. Do you see any problems with implementation?

saimaz:
Look ok.

PR #386, opened 2015-05-18
Changed symfony/symfony dependency to components

martiis:


saimaz:
@martiis are you sure we need all those components? Maybe some of them we can move to dev requirements?

martiis:
Actually, some of them could go in dev.

saimaz:
Already implemented in #394

PR #389, opened 2015-05-24
Custom repository class can be defined via Document annotation.

asev:
Custom repositories works for single typed repository getter only.

### DIFF
--- a/Annotation/Document.php
+++ b/Annotation/Document.php
@@ -32,6 +32,11 @@ final class Document
     /**
      * @var string
      */
+    public $repositoryClass;
+
+    /**
+     * @var string
+     */
     public $parent;
 
     /**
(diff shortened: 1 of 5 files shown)


saimaz:
Sorry @asev I don't get the point. Why do you need custom repository class? Repository class is not meant to be like doctrine's entity repository. Its just the pattern. The whole logic should be added to symfony services.

asev:
The main reason why I've implemented this - because many of Symfony developers (including me) are used to move most of query related logic to repositories. But actually, I see your point of moving this kind of logic to services.
Thank you @saimaz for quick response.

saimaz:
We could think about some custom repository for the custom queries, but it should be implemented in some proxy way.  I'm closing this issue until we will think about the proper implementation. The main goal is that a `Document` objects should remain unchanged.

PR #394, opened 2015-06-09
Changed SF minimum requirement to 2.7

saimaz:
There are already many places where we use functions added only in 2.6 (e.g. Mapping pass factory). I dont think is worth to keep support for 2.3 and lose all the greatness from symfony while 2.7 is new LTS release.


murnieza:
Looks reasonable for me.

PR #402, opened 2015-07-03
Add possibility to pass parameters to scripts

Dariusp:


saimaz:
For this case `ParametersTrait` was invented :). Just use `addParameter` function to add any needed params.

PR #404, opened 2015-07-10
Types can be specified while exporting index.

asev:


saimaz:
Merged in #405. Will be released in v1.0.0

Issue #407, opened 2015-07-15
Need custom field in Property annotation


tautrimas:
Sometimes yet unsupported field is needed in ES mapping, that ESB does not support. Custom fields can be easily supported by adding `raw` field to annotation:

```
@ES\Property(name="created_at", type="date", raw={"format":"basic_date"})
```

Currently, format field is unsupported, but could easily be defined using `raw` field, where all key/values in `raw` are passed directly to ES index mapping without validation. Name `raw` is just an example.

saimaz:
It will be introduced in v1.0.0, check out #405

saimaz:
Would be even more nice to be able to add any fields to annotation, because it confuses a lot which fields can be added natively and which to raw..

saimaz:
So, finaly it looks like it is solved. Please take a look at the new possible mapping document fields. https://github.com/ongr-io/ElasticsearchBundle/blob/48e3ab270f30c95dd2bb23ecd3904b9af7170fe9/Annotation/Property.php

The `options` variable was introduced, in witch we will use all custom fields, example: 
https://github.com/ongr-io/ElasticsearchBundle/blob/48e3ab270f30c95dd2bb23ecd3904b9af7170fe9/Tests/app/fixture/Acme/BarBundle/Document/Product.php#L29

Now you can add anything to the mapping and it will be validated by elasticsearch-php client.

PR #409, opened 2015-07-16
Add raw field to property annotation.

GrandLTU:
Allows setting properties that are not yet supported by ESB. #407


saimaz:
Merged in #405

PR #410, opened 2015-07-16
Suggester

GrandLTU:


saimaz:
Those are really great suggestions. We will definitely use this PR on 1.1.0 releae

saimaz:
I'm clossing it because we will do a cherry picks from some commits. Currently there is so many conflicts to merge it.

PR #422, opened 2015-08-20
Update TypeDropCommand.php

ddpkts:
updated command description


saimaz:
Type drop command is deprecated. No need to change that here. Look in to the #405 PR for all comming changes in v1.0.0 version.

PR #430, opened 2015-09-25
Elasticsearch events now works with Manager instead of Connection

asev:


asev:
Oh, I've just noticed that events are not used anymore. So, this pull request can be closed.

saimaz:
There will be events, they are removed just temporary to prepare a better integration for the 1.0 major version.

PR #442, opened 2015-10-26
simplified findBy function

### LINKED ISSUE #438: Use query_string query in the findBy() function to search results.

saimaz:
closes #438


tautrimas:
I believe, this change does not handle escaping, does it?

saimaz:
Escaping AFAIK is handled by elasticsearch-php. It validates json. If you have concers about comming value from the front end, then your application should handle the escaping.

tautrimas:
Of course JSON will be valid, no problems with that.

I am wondering if everything will be ok, if people start writing `$repository->findBy(['state' => 'OR'])` which stands for Oregon, USA.

With Term and Terms queries/filters there would be no such problem.

Maybe add a big fat warning about this?

saimaz:
Ok, got your point. Yes in these kind of situations there will be an error ....

saimaz:
Added an issue.

PR #462, opened 2015-11-05
Added curl configuration values.

ribb3r:
Added the ability to configure curl options (CURLOPT_CONNECTTIMEOUT, CURLOPT_TIMEOUT) via ongr_elasticsearch.yml.

```
...
    connections:
        default:
            hosts:
                - { host: 127.0.0.1:9200 }
            index_name: acme_index
            curl_options:
                connection_timeout: 2.0
                timeout: 2.0
...
```


saimaz:
@ribb3r is there something in 0.9 version you need specifically that we dont have in 1.0.alpha? I strongly reccomend you to upgrade a bundle version and do not use 0.9. dev-master is very stable release already. And most likely SDK wont change. Also in addition you could check out how it's working and suggest any bugs you notice.

ribb3r:
@saimaz Version 1.0.x requires Symfony 2.7 components. I'm currently using a Symfony 2.6 env. This is the reason why I have to use your 0.9 version.

saimaz:
This change will be only in 0.9. We are not gonna include it in upper versions. Hope @ribb3r it gonna help you until you upgrade sf2.

ribb3r:
@saimaz: Any reason why this won't be added to further versions? Do we have other options to handle connection timeouts?

saimaz:
It's too complicated to support whatever user might want. We will keep simplicity and provide most common use case for the client and if in any case it doesn't fit there will be an event where you could modify the creation of client and etc.

PR #474, opened 2015-11-30
Add missing datatypes

brammers:
Added missing ES data types.

https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html

### DIFF
--- a/Annotation/Property.php
+++ b/Annotation/Property.php
@@ -37,7 +37,7 @@ final class Property
      * @var string
      *
      * @Required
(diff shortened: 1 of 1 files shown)


murnieza:
Thanks for contribution. I am really concerned about `array` type. I think it should not be provided as any field can have one or more values by default (see https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html)

saimaz:
There is no array type in ES. @brammers can you please remove `array`?

brammers:
Sorry for the delay, yeah I see that now. I'll update the patch and fix the PSR-2 warning as well.

PR #476, opened 2015-11-30
Implement ArrayAccess

brammers:
Implemented ArrayAccess so you items using an array index.

### DIFF
--- a/Result/AbstractResultsIterator.php
+++ b/Result/AbstractResultsIterator.php
@@ -17,7 +17,7 @@ use ONGR\ElasticsearchBundle\Service\Repository;
 /**
  * Class AbstractResultsIterator.
  */
-abstract class AbstractResultsIterator implements \Countable, \Iterator
+abstract class AbstractResultsIterator implements \ArrayAccess, \Countable, \Iterator
 {
     /**
      * @var array Documents.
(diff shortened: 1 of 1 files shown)


mvar:
Please note that `key` used by iterator is nothing more than incremental number. In addition, it gets reset every **n** (scroll size) documents when using Elasticsearch's _scan and scroll_.

saimaz:
Our abstract results iterator will never has `ArrayAccess`. There is in the plans to provide an ability use your own iterator for the results. It might come in stable 1.0 or later in 1.1

brammers:
The problem I'm having is I can't read an item using an index on on a property. In the main Document I had a declaration as follows on the property.

    /**
     * @var Image[]
     *
     * @ES\Property(name="images", type="object", multiple=true, objectName="BundleName:Image")
     */
    protected $images;

Without the ArrayAccess implementation, I cannot access the items on the array by index in Twig.

Now based on your comments, would it make sense to re-do the patch on ObjectIterator instead?

saimaz:
It's interesting that you cannot iterate through images in the template. Because normaly everything should be fine. Can you please provide me more detailed example to recreate the problem you have?

brammers:
I'm not iterating - I'm getting the image object by index. The item is a document that I'm already iterating through.

In Twig I'm using 

{{ item.images[0].property }}

Or in Twig compiled PHP code

$code->images[0]->property;

However twig can't get the image at index 0 (Or any other position) because ObjectIterator does not implement ArrayAccess

Twig throws it out with "Impossible to access a key "%s" on an object of class "%s" that does not implement ArrayAccess interface"

https://github.com/twigphp/Twig/blob/604e94883bc1691071d9a30ff522808dc476394c/lib/Twig/Template.php#L484

saimaz:
Ok got it. You may try `first` twig helper. 

http://twig.sensiolabs.org/doc/filters/first.html

Anyway, I will investigate this problem, might be that we will add ArrayAccess to the object iterator (no promises) or will figure out a solution for this particular case atleast.

Issue #459, opened 2015-11-03
Converter throws an exception on empty objects


murnieza:
When document has field of type `object` and it's value is null the following exception is thrown:
>
ContextErrorException: Catchable Fatal Error: Argument 1 passed to ONGR\ElasticsearchBundle\Result\Converter::assignArrayToObject() must be of the type array, null given, called in /var/www/ongr-sandbox/vendor/ongr/elasticsearch-bundle/Result/Converter.php on line 107 and defined

Some kind of validation for empty objects has to be implemented.

ndinh215:
Hi Murnieza, in this case, should we return null with the 1st argurment's null value?

saimaz:
Converter has to create an empty `ObjectIterator` even there is no values for it.

ndinh215:
OK, thks Saimaz.

ndinh215:
I regenerated the exception case in the revision ~0.9, it did not show the exception. As Mantas's advice, I will inspect this issue in the dev-master revision. Thanks.

ndinh215:
I inspected and detected that the error was caused in the case of the object field's null value & non multiple. It falls into the else clause in the following code: 

```PHP
if ($aliases[$name]['multiple']) {                               
    $value = new ObjectIterator($this, $value, $aliases[$name]); 
} else {                                                         
    $value = $this->assignArrayToObject(                         
        $value,                                                  
        new $aliases[$name]['namespace'](),                      
        $aliases[$name]['aliases']                               
    );                                                           
}

ndinh215:
#489, it was fixed, please help me to review it. Thanks.

PR #489, opened 2015-12-08
Fix the exception of the case that an object field is empty & non multiple.

ndinh215:
The issue #459 was fixed. Please help me to review this patch.
Thanks.


saimaz:
@ndinh215 first of all we should try to not reduce code coverage with introduced changes. Can you provide a unit test case for the problem?

ndinh215:
Yup, I will create an unit testcase in Tests/Functional/Result/DocumentIteratorTest and alert you to review again. Thanks.

ndinh215:
Here is the test case for this issue: https://github.com/ndinh215/ElasticsearchBundle/commit/090b65f565cc36f55664fa7f08d1a871ab3ea723. Thanks.

patie:
i think this is problem again in v7, i have setter setImage(?Image $image); and i give error

`Product::setImage() must be an instance of App\\Elastic\\Document\\Advert\\Image or null, instance of ONGR\\ElasticsearchBundle\\Result\\ObjectIterator given`

PR #490, opened 2015-12-08
Check for object/nested notation added to DocumentParser

### LINKED ISSUE #481: Check if annotation is correct for object/nested
Say we are defining document and one field is `object`, and we are reference object's class as `objectName="AppBundle:Annexe"`.  

There is no check if `Annexe` object has correct annotation and PHP warning is thrown in case of wrong annotation: 
`Warning: array_replace_recursive(): Argument #2 is not an array` at this [line](https://github.com/ongr-io/ElasticsearchBundle/blob/65d6405ccb2cbd7fb79da9ef42745e9fdd39cd0f/Mapping/DocumentParser.php#L328).

There should be check if annotation is correct and detailed exception if it is not.

asev:
closes #481

### DIFF
--- a/Mapping/DocumentParser.php
+++ b/Mapping/DocumentParser.php
@@ -360,6 +360,12 @@ class DocumentParser
 
         $this->objects[$namespace] = $this->getRelationMapping(new \ReflectionClass($namespace));
 
+        if ($this->objects[$namespace] === null) {
+            throw new \LogicException(
+                sprintf('%s should have ES object or ES nested notation to be used as as object property.', $objectName)
+            );
+        }
+
         return $this->objects[$namespace];
     }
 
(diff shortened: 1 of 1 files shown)


mvar:
@asev please cover this use case with test.

Also, maybe it would be more clear to show message like `...should have @Object or @Nested anotation...`?

asev:
I wanted to add an unit test for this as well. But this change is at the deep private methods chain. It means, I should test public method `parse()`, and there is no unit test for DocumentParser at all. Is there a reason why it is not covered in unit tests? If no - I plan to invest some time and write some tests. But this shouldn't be related to this change.

Btw, I agree about the message. Already changed it. Thanks.

saimaz:
@mvar unit test is not the best approach to make sure for private method coverage improvements here. IMO much better to be write a functional test with not correct document tests. I'll take a look how to write such a test and if that is the right way.

PR #491, opened 2015-12-08
'is' getter now is supported for documents

### LINKED ISSUE #485: Add support for document getters with "is" prefix
It's common practice to have `setFoo()` and `isFoo()` methods for boolean variables. Currently only `get` prefix is supported.

Code reference: https://github.com/ongr-io/ElasticsearchBundle/blob/65d6405ccb2cbd7fb79da9ef42745e9fdd39cd0f/Mapping/DocumentParser.php#L187

asev:
closes #485

### DIFF
--- a/Mapping/DocumentParser.php
+++ b/Mapping/DocumentParser.php
@@ -182,23 +182,7 @@ class DocumentParser
                     case $property->isProtected():
                     case $property->isPrivate():
                         $propertyType = 'private';
-
-                        $camelCaseName = ucfirst(Caser::camel($name));
-                        if ($reflectionClass->hasMethod('get'.$camelCaseName)
-                            && $reflectionClass->hasMethod('set'.$camelCaseName)
-                        ) {
-                            $alias[$type->name]['methods'] = [
-                                'getter' => 'get'.$camelCaseName,
-                                'setter' => 'set'.$camelCaseName,
-                            ];
-                        } else {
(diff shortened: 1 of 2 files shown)


saimaz:
I have strong doubts about is'ser. Even if we will add a support for it, we have to check only on `boolean` fields.

mvar:
I think _is'sers_ are really useful. As @saimaz suggested, _is'ser_ should be checked only for `boolean` parameters.

asev:
IMHO issers make code more readable. To use isser only for boolean would need to implement complex logic.
Why is it bad to check getter/isser for all properties? [twig](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Template.php#L568) and [doctrine](https://github.com/doctrine/collections/blob/master/lib/Doctrine/Common/Collections/Expr/ClosureExpressionVisitor.php#L49) do that.

asev:
Ok, it was easier than I expected. So issers will be used for booleans only.

saimaz:
So before the merge we need to solve the question if we need to support both - isser and getter together.

mvar:
@saimaz are you talking about `boolean` fields only? If so, I think we should support both _issers_ and _getters_.

saimaz:
Ok. it was only a suggestions, but seems like it's completely enough to have a way to get and set the data. It's no matter if it's isser or getter.

PR #499, opened 2015-12-17
index creation with Elasticsearch 2.0

### LINKED ISSUE #493: Investigate _ttl behaviour with ES 2.0
We got complains that on ES 2.0 it is impossible to create mapping because of '_ttl' field.

[Official documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-ttl-field.html) states that "implementation is deprecated".

Full exception:
```json
{
   "error":{
      "root_cause":[
         {
            "type":"mapper_parsing_exception",
            "reason":"mapping [articles]"
         }
      ],
      "type":"mapper_parsing_exception",
      "reason":"mapping [articles]",
      "caused_by":{
         "type":"illegal_argument_exception",
         "reason":"Mapper for [_ttl] conflicts with existing mapping in other types:\n[mapper [_ttl] cannot be changed from type [string] to [long]]"
      }
   },
   "status":400
}
```

asev:
Also ManagerTest case fixed. Before this ttl was null and null is less than 1+.
closes #493

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -392,6 +392,12 @@ class Manager
             unset($this->indexSettings['body']['mappings']);
         }
 
+        if (isset($this->indexSettings['body']['mappings'])) {
+            foreach ($this->indexSettings['body']['mappings'] as $type => $mapping) {
+                unset($this->indexSettings['body']['mappings'][$type]['properties']['_ttl']);
+            }
+        }
+
         return $this->getClient()->indices()->create($this->indexSettings);
     }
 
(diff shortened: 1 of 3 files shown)


mvar:
This looks like a hack. What if we need to exclude other fields to? Actually we need! Currently we send to Elasticsearch not only `_ttl`, but also `_id`, `_score` and `_parent`. We shouldn't.

Maybe we could create separate annotation for such fields?

CC @saimaz @murnieza

saimaz:
This is definitelly a hack. No way we can do like this. Let's think again.

asev:
Thank you, @mvar. This is exactly why I created this pull request. I didn't like my solution, but it was all I got. It was more like discussion starter.

saimaz:
Closing this PR because it has to be implemented totaly different.

PR #500, opened 2015-12-17
Handling subdirectories in document folder

sidux:


saimaz:
May you provide a test case for this functionality?

mvar:
As we follow PSR-4 `getFileNamespace()` should use path to extract full namespace.

saimaz:
@sidux can you do a rebase with master, there were made some changes to ensure support for symfony 3.0, otherwise all tests will fail.

sidux:
@mvar yeah you are right but how am i supposed to know namespace root from document full path ?

saimaz:
@sidux yoou, this namespace grab doesn't look healthy.. Can't you get namespace from some `Reflection` ?

sidux:
@saimaz you can't use files with Reflection, the only thing i have there is the file path, but i can still read the composer.json and decode it to get the "autoload.psr-4" param so i can get the project root namespace.

mvar:
@sidux there is a way to get full namespace without parsing any files.

In [`getBundleDocumentPaths()`](https://github.com/sidux/ElasticsearchBundle/blob/b1f94d5b0ce2aec2694ba9b6c6e39b3514466298/Mapping/DocumentFinder.php#L83) you have first part of namespace (e.g., `Acme\DemoBundle`). And when you [recursively] get the list of documents in the `Document/` directory using `glob()`, you have the last part (plus `.php` extension). So full namespace would be like this (pseudo code):

```
BUNDLE_NAMESPACE + '\Document\' + (FILE_PATH_FROM_GLOB - EXTENSION)
```

You can change result format of `getBundleDocumentPaths()`. You need to return full namespace of each document instead of paths. You can also rename this method to `getBundleDocumentNamespaces()`. This will introduce BC break, but that's totally fine. These changes will be released with v1.0 (in early January).

Few notes:

- Wouldn't it be nicer to use `RecursiveDirectoryIterator` class instead of `glob()` function?
- It looks like you did git merge instead of rebase (your branch duplicates existing commits)

Thanks for contributing!

mvar:
@sidux what's the status of this PR?

btw, happy New Year!

PR #527, opened 2016-01-04
Introduce Collection class as a holder for multiple objects

### LINKED ISSUE #525: Impossible to append new object to object iterator
When user has document with "multiple object" field, there is no way to append new object to that field in case you read document from ES.

mvar:
This PR adds requirement to use `Collection` for "multiple objects" properties. `ObjectIterator` now extends `Collection`.

Not sure about the naming and namespaces.

Closes #525

### DIFF
--- a/Collection.php
+++ b/Collection.php
@@ -1,0 +1,117 @@
+<?php
+
+/*
+ * This file is part of the ONGR package.
+ *
+ * (c) NFQ Technologies UAB <info@nfq.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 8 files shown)


saimaz:
`Collection` as a name is too abstract, IMO at least it has to be named `ObjectCollection` or something similar.

saimaz:
Also the docs about how to define object/nested is missing. I mean we have to write in the docs that there is necessary to add a collection in a document constructor.

mvar:
Name left as it is. It may sound abstract, but it actually is quite abstract. `Collection` is basically array wrapper. We use it for single case, but it could be easily used for many other purposes.

Also updated docs.

Issue #423, opened 2015-09-16
Add --try flag to ongr:es:index:create command


martiis:
Then if index already exists exception should not be thrown, but a clear message outputed saying that index already exists or smth..

kazgurs:
+1

mvar:
Please give right name for this option. `--try` is misleading. As far as I understand the expected result is to create index or silently ignore exception if it is already created.

BTW, Doctrine has similar flag on `database:create` command. It's named `--if-not-exists`. Not sure about such name, but at least it's not misleading.

ndinh215:
Yes, this is also my concern. "--if-not-exists" is correct option in Doctrine. Here, we just show error message instead of exception in the case of existing a certain index.

saimaz:
Agree with @mvar `--try` doesn't say anything. `--if-not-exists` is more correct option name.

ndinh215:
I fixed it, please review the PR https://github.com/ongr-io/ElasticsearchBundle/pull/535. Thanks.

mvar:
From now on we have `--if-not-exists` flag to control if command should check for index existence or not.

PR #535, opened 2016-01-05
Add the option '--if-not-exists' in the command 'ongr:es:index:create'

ndinh215:
I fixed the issue #423 . Please review again. Thanks.

### DIFF
--- a/Command/IndexCreateCommand.php
+++ b/Command/IndexCreateCommand.php
@@ -38,7 +38,13 @@ class IndexCreateCommand extends AbstractManagerAwareCommand
                 InputOption::VALUE_NONE,
                 'If the time suffix is used, its nice to create an alias to the configured index name.'
             )
-            ->addOption('no-mapping', null, InputOption::VALUE_NONE, 'Do not include mapping');
+            ->addOption('no-mapping', 'nm', InputOption::VALUE_NONE, 'Do not include mapping')
+            ->addOption(
+                'if-not-exists',
+                null,
+                InputOption::VALUE_NONE,
+                'Don\'t trigger an error, when the index already exists'
+            );
     }
 
     /**
@@ -53,6 +59,17 @@ class IndexCreateCommand extends AbstractManagerAwareCommand
             /** @var IndexSuffixFinder $finder */
             $finder = $this->getContainer()->get('es.client.index_suffix_finder');
             $finder->setNextFreeIndex($manager);
+        }
+
+        if ($input->getOption('if-not-exists') && $manager->indexExists()) {
+            $output->writeln(
+                sprintf(
+                    '<info>Index `<comment>%s</comment>` already exists in `<comment>%s</comment>` manager.</info>',
+                    $manager->getIndexName(),
+                    $input->getOption('manager')
+                )
+            );
+            return;
         }
 
         $manager->createIndex($input->getOption('no-mapping'));
--- a/Command/IndexCreateCommand.php
+++ b/Command/IndexCreateCommand.php
@@ -41,7 +41,7 @@ class IndexCreateCommand extends AbstractManagerAwareCommand
             ->addOption('no-mapping', 'nm', InputOption::VALUE_NONE, 'Do not include mapping')
             ->addOption(
                 'if-not-exists',
-                null,
+                'ine',
                 InputOption::VALUE_NONE,
                 'Don\'t trigger an error, when the index already exists'
             );
--- a/Command/IndexCreateCommand.php
+++ b/Command/IndexCreateCommand.php
@@ -41,7 +41,7 @@ class IndexCreateCommand extends AbstractManagerAwareCommand
             ->addOption('no-mapping', 'nm', InputOption::VALUE_NONE, 'Do not include mapping')
             ->addOption(
                 'if-not-exists',
-                'ine',
+                null,
                 InputOption::VALUE_NONE,
                 'Don\'t trigger an error, when the index already exists'
             );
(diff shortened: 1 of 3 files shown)


Command/IndexCreateCommand.php
```diff
@@ -38,7 +38,13 @@ protected function configure()
                 InputOption::VALUE_NONE,
                 'If the time suffix is used, its nice to create an alias to the configured index name.'
             )
-            ->addOption('no-mapping', 'nm', InputOption::VALUE_NONE, 'Do not include mapping');
+            ->addOption('no-mapping', 'nm', InputOption::VALUE_NONE, 'Do not include mapping')
+            ->addOption(
+                'if-not-exists',
+                'ine',
```
murnieza: I think that Symfony does not support multi-letter input option shortcuts. I see that we have the same with other options

ndinh215: http://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
With this standard, it could understand wrong the options.

Command/IndexCreateCommand.php
```diff
@@ -55,6 +61,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
+        if ($input->getOption('if-not-exists') && $manager->indexExists()) {
+            $output->writeln(
+                sprintf(
+                    '<info>Index `<comment>%s</comment>` already exists in `<comment>%s</comment>` manager.</info>',
+                    $manager->getIndexName(),
+                    $input->getOption('manager')
+                )
+            );
+            return;
```
saimaz: I'm not sure 100% but probably better would be change to: `return 1;`.

ndinh215: You are correct. As Doctrine's code about this processing, it will be 
```
return $error ? 1 : 0;
```
whatever success or error.

Tests/Functional/Command/CreateIndexCommandTest.php
```diff
@@ -64,6 +64,37 @@ public function testExecute($managerName, $arguments, $options)
...
     /**
+     * Tests creating index in case of existing this index. Configuration from tests yaml.
+     *
+     * @param string $managerName
+     * @param array  $arguments
+     * @param array  $options
+     *
+     * @dataProvider getTestExecuteData
+     */
+    public function testExecuteWithIndexExistence($managerName, $arguments, $options)
```
saimaz: You should not use different test case data provider and try to behave inside test case to match expectations. To test `--if-not-exists` please write separate test case, the data provider is not necessary.

Tests/Functional/Command/CreateIndexCommandTest.php
```diff
@@ -64,6 +64,37 @@ public function testExecute($managerName, $arguments, $options)
...
+    {
+        $manager = $this->getManager($managerName);
+
+        if (!$manager->indexExists()) {
+            $manager->createIndex();
+        }
+
+        try {
+            $arguments['--if-not-exists'] = null;
```
saimaz: it's actually an option, not argument.

Tests/Functional/Command/CreateIndexCommandTest.php
```diff
@@ -64,6 +64,37 @@ public function testExecute($managerName, $arguments, $options)
...
+     */
+    public function testExecuteWithIndexExistence($managerName, $arguments, $options)
+    {
+        $manager = $this->getManager($managerName);
+
+        if (!$manager->indexExists()) {
+            $manager->createIndex();
+        }
+
+        try {
```
saimaz: When you will write test case, we do not need exception catch here.

ndinh215:
Thanks for your points, @saimaz . With this patch, I removed ```try catch``` and added asserts for testing result which is returned from the execution and the output string. And also, I refactored the code a bit for the purpose of retrieving ```command tester``` via the method:
```
protected function getCommandTester($commandName)
    {
        $app = new Application();
        $app->add($this->getCreateCommand());

        $command = $app->find($commandName);
        $commandTester = new CommandTester($command);

        return $commandTester;
    }
```
About the data provider, I think this testcase may still use the same data provider. It could iterate all execute data and test this case with the option 'if-not-exists'.

Tests/Functional/Command/CreateIndexCommandTest.php
```diff
@@ -64,6 +64,49 @@ public function testExecute($managerName, $arguments, $options)
...
+     *
+     * @dataProvider getTestExecuteData
+     */
+    public function testExecuteWhenIndexExists($managerName, $arguments, $options)
+    {
+        $manager = $this->getManager($managerName);
+
+        // Try to create index in prior
+        if (!$manager->indexExists()) {
+            $manager->createIndex();
```
mvar: Index is automatically created when you call `$this->getManager()`.

Tests/Functional/Command/CreateIndexCommandTest.php
```diff
@@ -64,6 +64,49 @@ public function testExecute($managerName, $arguments, $options)
...
+        $expectedOutput = sprintf(
+            'Index `%s` already exists in `%s` manager.',
+            $manager->getIndexName(),
+            $managerName
+        );
+
+        // Test if the command output matches the expected output or not
+        $this->assertStringMatchesFormat($expectedOutput . '%a', $commandTester->getDisplay());
+
+        $manager->dropIndex();
```
mvar: Created indexes are automatically removed after each test case.

ndinh215:
Closes #423

ndinh215:
I will create new PR with better form for this issue.

PR #543, opened 2016-01-06
[Bugfix] Dont set filtered query into endpoint during normalization

martiis:
If same search happens to be executed twice, then in second query we would see filtered query within a filtered query with same filters.

This issue also slows down query execution time.

** EDITED **

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,6 @@ install:
   - "echo 'script.disable_dynamic: false' | sudo tee -a /etc/elasticsearch/elasticsearch.yml"
   - "echo 'index.warmer.enabled: true' | sudo tee -a /etc/elasticsearch/elasticsearch.yml"
   - sudo service elasticsearch restart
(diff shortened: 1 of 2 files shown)


mvar:
@martiis could you add test case for this issue? As far as I understand it's enough to normalize query twice and check if results are equal, right?

> This also slows down query execution time.

The issue or the solution? :) I guess the issue. Please be more specific next time.

martiis:
1. Yes
2. Ok ;)

saimaz:
@martiis can you open the same PR to the DSL library? Of course, we will merge and release the `0.*` but it would be helpful also in a new version ;)

saimaz:
@martiis could you remove python docs building part in travis? Then I could merge this PR.

martiis:
I've removed it.

saimaz:
The tests are still failing. I'm not sure that for 0.10 version this change is actual. If I'm wrong please reopen this PR with fixed test case.

Issue #382, opened 2015-05-15
Index import command should be able to import gzipped files


tautrimas:
Currently, importing is possible only on plain .json files. It should be very easy to enable gzipped file import by using this snippet:

```php
$fileHandler = @fopen(((substr($this->filename, -3) == '.gz') ? 'compress.zlib://' : '') . $this->filename, 'r');
```

https://github.com/ongr-io/ElasticsearchBundle/blob/e95531a69430f23bb886a39a221b449a69433f72/Service/Json/JsonReader.php#L99

saimaz:
Any chance to get PR for this?

ndinh215:
Do we need an option to distinguish between a normal file with a compressed Json file? For example, in MongoDB, they use the option '--jsonArray' to differently behave with traditional Json file and json objects array.
```
mongoimport --db <db-name> --collection <coll-name> --type json --file data.json
--jsonArray
```
How do you think? @saimaz @mvar @murnieza

saimaz:
Yes you are absolutely correct. The provided example by cutting 3 last letters from the path is a wrong way. We should introduce option `--gzip` and then use `compress.zlib://` stream. If user messes up with the file extensions that his problem not our's.

ndinh215:
OK, got it. Thanks @saimaz .

ndinh215:
Closes #382.

PR #549, opened 2016-01-08
Support import gzip json file by command

ndinh215:
I created new PR. Please review this implemention. Thanks.


Command/IndexImportCommand.php
```diff
@@ -53,13 +59,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        }
+
         /** @var ImportService $importService */
         $importService = $this->getContainer()->get('es.import');
         $importService->importIndex(
             $manager,
             $input->getArgument('filename'),
             $output,
-            $input->getOption('bulk-size')
+            $input->getOption('bulk-size'),
```
murnieza: `bulk-size` could also be in `$options`.

ndinh215:
closing #382

murnieza:
That's just a thought. Maybe we should leave it as a separated argument as it is required. Opinions?

Command/IndexImportCommand.php
```diff
@@ -53,13 +59,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $manager = $this->getManager($input->getOption('manager'));
 
+        $options = [];
+        if ($input->getOption('gzip')) {
+            $options['gzip'] = true;
```
mvar: You should set it to `null` with current implementation. With `true`/`false` it behaves the same, as with any other value.

ndinh215:
@saimaz  & @mvar , how do you think? About me, its rather appropriate.

Command/IndexImportCommand.php
ndinh215: Yup, regardless of the option's value, I only check the existence of the key 'gzip'. Your idea is nice. I will fix it. Thanks.

Command/IndexImportCommand.php
saimaz: Actually, it must be in `$options`, that's why this variable was introduced, to collect all options in a single place.

Service/Json/JsonReader.php
```diff
@@ -66,18 +66,26 @@ class JsonReader implements \Countable, \Iterator
...
      * Constructor.
      *
      * @param Manager $manager
      * @param string  $filename
+     * @param array   $options
      * @param bool    $convertDocuments
+     *
      */
-    public function __construct($manager, $filename, $convertDocuments = true)
+    public function __construct($manager, $filename, $options, $convertDocuments = true)
```
saimaz: $convertDocuments might be good to put in `$options` too. Ideas? @ndinh215 @mvar ??

Command/IndexImportCommand.php
ndinh215: Total agree, @saimaz

Service/Json/JsonReader.php
ndinh215: I think the options here is the options of the command. Thus, ```$convertDocuments``` should not be put into the options.

saimaz:
I'm looking at the JsonReader and I think that maybe much better would be to include in it `ParametersBagTrait`. Then all options has to be set via setters and we won't need so long constructor. Also it's very difficult to maintain constructor, we simply cannot change the structure.

Service/Json/JsonReader.php
saimaz: I took a bit deeper look and don't think we need this argument. Nothing in the whole bundle uses this option. And by current logic nothing will. Propose to remove `$convertDocuments`. And then remove `convertDocument()` function.

ndinh215: OK, got it, @saimaz . Thanks.

ndinh215:
I removed the function convertDocument and refactored again. Please review it. Thanks.

PR #555, opened 2016-01-19
Refactor aggregations result handling

### LINKED ISSUE #554: Write docs:  how to use aggregations iterator

mvar:
Main changes:

### Quicker properties access:

Now you can directly access aggregation properties using array access syntax.

Before:

```php
$result->getAggregation('range_agg')->getValue()['min'];
```

After:

```php
$result->getAggregation('range_agg')['min'];
```

### Logically separated buckets

From now on all aggregations are returned as instance of `ValueAggregation` (previously bucketed aggregations were returned as iterator of buckets). This means all aggregation properties can be reached the same way as with metric aggregations. You still can iterate over buckets by using aggregation itself as iterator. You can get array of buckets by calling `getBuckets()`.

Before:

```php
$buckets = $result->getAggregation($name);
```

After:

```php
$fooValue = $result->getAggregation($name)['foo_value'];
$buckets = $result->getAggregation($name)->getBuckets();
```

--

Closes #554


saimaz:
Lets release ElasticsearchDSL 2.0 and this will be a way to go.

mvar:
Updated code to work with DSL 2.0 (`master` branch currently).

This PR is ready to merge, but before merging we should discuss about naming. I don't like neither `ValueAggregation` nor `AggregationValue`.

saimaz:
AggregationValue seems a good choice, otherwise the name will be inaccurate.

PR #563, opened 2016-02-01
Remove "delete by query" logic

### LINKED ISSUE #553: DeleteByQuery was removed in Elasticsearch 2.0
https://www.elastic.co/guide/en/elasticsearch/reference/1.6/docs-delete-by-query.html

IMO we should leave the function but instead to call deleteByQuery maybe we should form search query for ID's and then create bulk remove query?

Ideas?

mvar:
Closes #553

### DIFF
--- a/Service/Repository.php
+++ b/Service/Repository.php
@@ -201,27 +201,6 @@ class Repository
     }
 
     /**
-     * Delete by query.
-     *
-     * @param Search $search
-     *
-     * @return array
-     */
-    public function deleteByQuery(Search $search)
-    {
-        $params = [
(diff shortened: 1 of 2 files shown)


mhaeussler-magari:
@saimaz @mvar There is a plugin from elastic which adds this functionality very easy: https://www.elastic.co/guide/en/elasticsearch/plugins/2.2/plugins-delete-by-query.html

Would be nice if the function would be available as long there is no own solution by ongr.

saimaz:
I will reopen an issue for discussion. We could consider to put it back in 1.1 or 1.2 version. Right now it is too late to bring it back until 1.0 release.

PR #565, opened 2016-02-03
Added missing 'useAttributeAsKey' to Configuration.php

christianbader:
Overwritten properties as 'managers' and/or 'connections' in elasticseach_config.yml are not merged as expected when 'useAttributeAsKey' is not defined in 'DependencyInjection/Configuration.php'.

Current result of merge:

````
ongr_elasticsearch:
    connections:
…
    managers:
        registrations: { connection: registrations }
        0: { connection: registrations_overwritten }
```

instead of expected:

````
ongr_elasticsearch:
    connections:
…
    managers:
        registrations: { connection: registrations_overwritten }
```


saimaz:
Thanks for the fix.

PR #569, opened 2016-02-25
getAggregations in DocumentIterator

juliensantos87:
allow to get all Aggregations from DocumentIterator

### DIFF
--- a/Result/DocumentIterator.php
+++ b/Result/DocumentIterator.php
@@ -38,6 +38,22 @@ class DocumentIterator extends AbstractResultsIterator
     }
 
     /**
+     * Returns aggregations.
+     *
+     * @return array
+     */
+    public function getAggregations()
+    {
+        $aggregations = array();
+
+        foreach ($this->aggregations as $key => $aggregation) {
+            $aggregations[$key] = $this->getAggregation($key);
+        }
+
+        return $aggregations;
+    }
+
+    /**
      * Get a specific aggregation by name. It fetches from the top level only.
      *
      * @param string $name


saimaz:
Is there any particular reason to have this? Because this kind of way only confuse me.. IMO you should request for only a specific agg which you set.

saimaz:
Maybe @mvar has some other thoughts

juliensantos87:
I was using it from V0.10 https://github.com/ongr-io/ElasticsearchBundle/blob/v0.10/Result/DocumentIterator.php#L126

It's actually useful in my opinion because It's allow to loop over all aggregations even if the part of my application doesn't know about aggregation name.

juliensantos87:
And it is in RawIterator https://github.com/ongr-io/ElasticsearchBundle/blob/master/Result/RawIterator.php#L24 

Maybe the Abstract should not be protected ? https://github.com/ongr-io/ElasticsearchBundle/blob/master/Result/AbstractResultsIterator.php#L115

saimaz:
We changed a lot with aggregations in 1.0-dev. Well I will take a look again to it. I agree that it would be usefull to foreach through all aggs. I will check if that doesn't make any conflicts or logic faults. If not I will add this function. 

btw, to have test case would be awesome, may you add it?

juliensantos87:
test case won't run on my computer. I will add some if I find time to figure it out.

saimaz:
I wonder are there any difficulties to run tests for this bundle specifically?

juliensantos87:
I don't know tests are running well on Travis.

I juste try quickly with this:
  vendor/bin/phpunit Tests/Functional/Result/DocumentIteratorTest.php 

And I got that:

1) ONGR\ElasticsearchBundle\Tests\Functional\Result\DocumentIteratorTest::testCurrentWithEmptyIterator
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: You cannot set service "service_container".

/Projects/ElasticsearchBundle/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Container.php:169
/Projects/ElasticsearchBundle/Tests/app/cache/test/appTestProjectContainer.php:23
/Projects/ElasticsearchBundle/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:485
/Projects/ElasticsearchBundle/Tests/app/AppKernel.php:71
/Projects/ElasticsearchBundle/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:117
/Projects/ElasticsearchBundle/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php:146
/Projects/ElasticsearchBundle/Test/AbstractElasticsearchTestCase.php:211
/Projects/ElasticsearchBundle/Test/AbstractElasticsearchTestCase.php:236
/Projects/ElasticsearchBundle/Test/AbstractElasticsearchTestCase.php:66

PR #572, opened 2016-03-01
Makes the Converter respect stange date formats

tim-massey:
The dates I store in elastic are formatted as follows:
 
```php
/**
  * @ES\Property(type="date", options={"format": "Y-M-D'T'HH:mm:ss"})
  */
  public $start;
```

Before this change they are Converted as `false`.

After this change, they are correctly converted into `\DateTime` objects.


saimaz:
Can you cover the bug you found with the test case?

PR #584, opened 2016-03-08
Implemented getDocumentSort() and wrote a test for it

### LINKED ISSUE #573: Get document sort field results
Hello, I would like to get the document's sort field in results, but so far I believe there isn't a method to do this. It is returning the sort field as an array in the protected documents property on search results but I can't figure out how to get the value. 

https://www.elastic.co/guide/en/elasticsearch/guide/current/sorting-by-distance.html

something like:
```
            foreach ($results as $doc) {
                $distance = $doc->getSort();
            }
```

einorler:
Also added documentation for the method
closes #573

### DIFF
--- a/Result/AbstractResultsIterator.php
+++ b/Result/AbstractResultsIterator.php
@@ -310,6 +310,24 @@ abstract class AbstractResultsIterator implements \Countable, \Iterator
     }
 
     /**
+    * Returns sort of current hit.
+    *
+    * @return mixed
+    */
+    public function getDocumentSort()
+    {
+        if (!$this->valid()) {
+            throw new \LogicException('Document sort is available only while iterating over results.');
+        }
+
+        if (!isset($this->documents[$this->key]['sort'])) {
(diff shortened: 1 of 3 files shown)


Result/AbstractResultsIterator.php
```diff
@@ -310,6 +310,24 @@ public function getDocumentScore()
     }
 
     /**
+    * Returns sort of current hit.
+    *
+    * @return array
+    */
+    public function getDocumentSort()
```
saimaz: I'm not sure the name is correct. `getSort()` as provided in issue looks much better.

mvar: I guess this follows `getDocumentScore()`. Later we will add getter for `id` too, so I think it would be nice to have single pattern.

saimaz:
@einorler Please add some notes about usage in docs. `getDocumentSort()` as @mvar said makes perfect sense.

PR #585, opened 2016-03-09
implemented type casting for float and int in Converter

### LINKED ISSUE #571: Elasticsearch cannot handle float value provided as a string
e.g. json import creates a document array where all values are strings, if in the mapping are float value, the document field won't be inserted due error. In the converter, there has to be check if value by the mapping is float or int, it has to be changed to according value type. 

https://github.com/ongr-io/ElasticsearchBundle/blob/77e264a55c97608765acf6935b86e9fdb9507d85/Result/Converter.php#L139

einorler:
Also wrote a test for it
closes #571


Tests/Functional/Result/PersistObjectsTest.php
```diff
@@ -91,4 +92,33 @@ public function testAppendObject()
...
+     * when objects are created with strings
+     */
+    public function testType()
+    {
+        $manager = $this->getManager();
+
+        $product = new Product();
+        $product->id = 'foo';
+        $product->title = 'Test Document';
+        $product->price = '8.95';
```
saimaz: What if a float value is 8,95 (comma instead a doc). This depends on locale settings, but still. @mvar do you think it's worth to consider?

einorler:
The possible issue has been addressed, now the converter converts both 0.00 and 0,00 to float

PR #589, opened 2016-03-15
implemented retrievement of id after commit()

### LINKED ISSUE #484: Set document ID after commit to ES
```php
$document = new CustomDocument();

$manager->persist($document);
$manager->commit();

var_dump($document->getId());
```

Currently this code returns `NULL`. It would be nice to get generated ID.

einorler:
closes #484

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -81,6 +81,14 @@ class Manager
     private $commitMode = 'refresh';
 
     /**
+     * Reference to the persisted objects that need to be given an ID that returns from
+     * elasticsearch
+     *
+     * @var array
+     */
+    private $persistedObjects = [];
+
+    /**
      * The size that defines after how much document inserts call commit function.
      *
      * @var int
@@ -274,6 +282,9 @@ class Manager
         $type = $this->getMetadataCollector()->getDocumentType(get_class($document));
 
         $this->bulk('index', $type, $documentArray);
+        if (!isset($document->id)) {
+            $this->persistedObjects[] = $document;
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Service/Manager.php
```diff
@@ -321,6 +332,38 @@ public function refresh(array $params = [])
...
+            if (isset($query['index']) && !isset($query['index']['_id'])) {
+                $indexing[] = $number / 2;
+            }
+        }
+
+        if (isset($bulkQueries['body'][0]['index'])) {
+            if (isset($this->persistedObjects)) {
+                $i = 0;
+                foreach ($this->persistedObjects as $document) {
+                    $document->id = $bulkResponse['items'][$indexing[$i]]['create']['_id'];
```
mvar: You can not hardcode `id` property here. User is free to choose which property of a document he wants to use for Elasticsearch `_id` value (using `@Id` annotation).

You could get actual property from metadata collector.

einorler:
Thanks for that, I've taken it into account in the latest changes

Service/Manager.php
```diff
@@ -321,6 +332,44 @@ public function refresh(array $params = [])
...
+        if (isset($bulkQueries['body'][0]['index'])) {
+            if (isset($this->persistedObjects)) {
+                $i = 0;
+                foreach ($this->persistedObjects as $document) {
+                    $class = get_class($document);
+                    $mapping = $this->metadataCollector->getMapping($class);
+
+                    if (isset( $mapping['aliases']['_id'])) {
+                        $id_property = $mapping['aliases']['_id']['propertyName'];
+                        $document->$id_property = $bulkResponse['items'][$indexing[$i]]['create']['_id'];
```
mvar: This property can be private :)

saimaz:
It's been remastered in #660 with grater performance.

PR #597, opened 2016-03-18
Rename MetaField interface

tompoc:
Renamed `MetaField` interface to `MetaFieldInterface`.


saimaz:
We will follow Symfony practise in this way. Maybe we could consider change interface name in 2.0...

PR #604, opened 2016-03-29
Check if Document dir exists

### LINKED ISSUE #603: Add check if Document dir exists
If there is defined bundle in mapping without `Document` directory Elasticsearch bundle throws exception. There should be a check if that directory exists before scan.

tompoc:
Check for document directory before mapping.

Closes #603


saimaz:
This PR should be opened for 1.0 branch

PR #612, opened 2016-03-31
Force bulk commit

tompoc:
Force commit to Elasticsearch on kernel terminate event for each manager


saimaz:
Please add some note to the docs as well

PR #622, opened 2016-05-06
Cache engine class parameter set

kamillo1888:
It is very important, some projects not need to cache results in file, for example our project has Memcached storage.


kamdz:
+1

kwojc:
+1

pawelkolanowski:
+2

saimaz:
Hello guys,

To solve your problem is very easy. Just overwrite `es.cache_engine` service with the same name but different class. In your case use Memcache driver. That's why we are using doctrine cache engine, where is easy to swith to any of the cache sources. 

I'm not gonna merge this PR because symfony doesn't define class namespaces as a params. The better practice is to overwrite service. It's much more stable to maintain backward compatibility accros major versions. 

I'm glad to hear that this bundle is usefull for you ;).

PR #628, opened 2016-05-13
[wip] Implemented initial Field annotation

jaytaph:
Initial release. It seems to be working, but it needs unit-tests and I'd reckon there are many things that i've overlooked.


saimaz:
Yes it has to be tested before merge. I will try to play with your branch to see if that works well with the rest of the bundle. 

Thanks for PR.

PR #633, opened 2016-06-09
made es manager allways refresh and flush only the current index

### LINKED ISSUE #629: Refresh on all indexes increases server load ~3x
I modified `refresh` method to look like this in 0.10 branch.

```php
    public function refresh()
    {
        $this->getClient()->indices()->refresh(['index' => $this->getIndexName()]);
    }
```

This decreased load from our ES servers about 3 times during heavy indexing and searching. Probably due to the fact, that up until now, all `_refresh` commands were performed on ALL indexes. This needs to be fixed for `_flush` probably as well.

einorler:
closes #629

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -297,6 +297,7 @@ class Manager
      */
     public function flush(array $params = [])
     {
+        $params = array_merge($params, ['index' => $this->getIndexName()]);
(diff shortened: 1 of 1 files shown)


saimaz:
@einorler please reopen this to 1.0

PR #634, opened 2016-06-09
fixed an issue with bad serialization

### LINKED ISSUE #624: Query given in profiler is not executable
Snippet of the query:

```json
"reverse": {
    "reverse_nested": [],
    "aggregations": {
        "foo": {
            "terms": {
               "field": "id"
            }
       },
```

Error:

```
"type": "search_parse_exception",
"reason": "Expected [START_OBJECT] under [reverse_nested], but got a [START_ARRAY] in [reverse]",
```

Somehow `reverse_nested` is printed as array instead of object (probably all other empty objects too). Query execution using ESB works, just query given in the profiler is broken.

I guess problem comes from this line:
https://github.com/ongr-io/ElasticsearchBundle/blob/b305614610c264e4da6b3097b536c383e15d6add/Profiler/ElasticsearchProfiler.php#L176

einorler:
I've looked at serializers in elasticsearch-php and they're not doing anything fancy at all, this is the solution that would fix the problem in question

closes #624

### DIFF
--- a/Profiler/ElasticsearchProfiler.php
+++ b/Profiler/ElasticsearchProfiler.php
@@ -173,7 +173,7 @@ class ElasticsearchProfiler implements DataCollectorInterface
     private function addQuery($route, $record, $queryBody)
     {
         parse_str(parse_url($record['context']['uri'], PHP_URL_QUERY), $httpParameters);
-        $body = json_decode(trim($queryBody, " '\r\t\n"), true);
+        $body = json_decode(trim($queryBody, " '\r\t\n"));
         $this->queries[$route][] = array_merge(
             [
                 'body' => $body !== null ? json_encode($body, JSON_PRETTY_PRINT) : '',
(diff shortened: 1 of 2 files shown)


Profiler/ElasticsearchProfiler.php
```diff
@@ -176,7 +176,7 @@ private function addQuery($route, $record, $queryBody)
         $body = json_decode(trim($queryBody, " '\r\t\n"), true);
         $this->queries[$route][] = array_merge(
             [
-                'body' => $body !== null ? json_encode($body, JSON_PRETTY_PRINT) : '',
+                'body' => $body !== null ? str_replace(': []', ': {}', json_encode($body, JSON_PRETTY_PRINT)) : '',
```
mvar: Doesn't this destroy real empty arrays?

einorler:
Yes, I was thinking of that too and it does. Any suggestions on how to distinguish between them?

saimaz:
This looks like too complex solution. How about if we could prettify JSON output directly in the browser?

saimaz:
btw fix IMO should be applied to `1.0`

einorler:
Closed due to a wrong branch

PR #638, opened 2016-06-13
made es manager allways refresh and flush only the current index

### LINKED ISSUE #629: Refresh on all indexes increases server load ~3x

einorler:
closes #629

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -298,7 +298,7 @@ class Manager
      */
     public function flush(array $params = [])
     {
-        return $this->client->indices()->flush($params);
+        return $this->client->indices()->flush(array_merge(['index' => $this->getIndexName()], $params));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Service/Manager.php
```diff
@@ -298,6 +298,7 @@ public function remove($document)
      */
     public function flush(array $params = [])
```
saimaz: Please change to this:

```php
public function flush(array $params = [])
{
    return $this->client->indices()->flush(array_merge(['index' => $this->getIndexName()], $params));
}
```

Also change the same way function below.

PR #639, opened 2016-06-13
implemented returning id in RESULTS_ARRAY

### LINKED ISSUE #468: Include document id and other core fields within Repository::RESULTS_ARRAY response
Currently only _source content is returned. Would be very handy to get also at least a document id.

einorler:
closes #468

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -590,6 +590,7 @@ class Manager
     private function convertToNormalizedArray($data)
     {
         if (array_key_exists('_source', $data)) {
+            $data['_source']['_id'] = $data['_id'];
             return $data['_source'];
         }
 
@@ -597,6 +598,7 @@ class Manager
 
         if (isset($data['hits']['hits'][0]['_source'])) {
             foreach ($data['hits']['hits'] as $item) {
+                $item['_source']['_id'] = $item['_id'];
                 $output[] = $item['_source'];
             }
         } elseif (isset($data['hits']['hits'][0]['fields'])) {
--- a/Tests/Unit/Service/ManagerTest.php
+++ b/Tests/Unit/Service/ManagerTest.php
@@ -303,6 +303,7 @@ class ManagerTest extends \PHPUnit_Framework_TestCase
         $client = $configuration[1];
         $search = $configuration[2];
         $array = [
+            '_id' => '15',
             '_source' => [
                 [
                     'price' => 1,
@@ -312,7 +313,7 @@ class ManagerTest extends \PHPUnit_Framework_TestCase
         ];
         $client->expects($this->any())->method('search')->willReturn($array);
         $result = $manager->execute(['product'], $search, 'array');
-        $this->assertEquals($result, $array['_source']);
+        $this->assertEquals($result, array_merge($array['_source'], ['_id' => $array['_id']]));
     }
 
     /**


Service/Manager.php
```diff
@@ -589,15 +589,20 @@ private function parseResult($raw, $resultsType, $scrollDuration = null)
      */
     private function convertToNormalizedArray($data)
     {
-        if (array_key_exists('_source', $data)) {
-            return $data['_source'];
+        if (array_key_exists('_source', $data) && array_key_exists('_id', $data)) {
```
saimaz: `_id` field always comes back in a response. No need to check for that. `_source` is checked in case of fields might be set in the query then `_source` won't come back.

Service/Manager.php
```diff
@@ -589,15 +589,20 @@ private function parseResult($raw, $resultsType, $scrollDuration = null)
...
         }
 
         $output = [];
 
         if (isset($data['hits']['hits'][0]['_source'])) {
             foreach ($data['hits']['hits'] as $item) {
-                $output[] = $item['_source'];
+                $output[$item['_id']] = [
+                    '_type'  => $item['_type'],
+                    '_index'  => $item['_index'],
```
saimaz: Index will always be the same. Repository class works with 1 manager at the time.

Service/Manager.php
```diff
@@ -589,15 +589,20 @@ private function parseResult($raw, $resultsType, $scrollDuration = null)
...
+            return [$data['_id'] => $data['_source']];
         }
 
         $output = [];
 
         if (isset($data['hits']['hits'][0]['_source'])) {
             foreach ($data['hits']['hits'] as $item) {
-                $output[] = $item['_source'];
+                $output[$item['_id']] = [
+                    '_type'  => $item['_type'],
```
saimaz: The same as in the comment below about index.

Service/Manager.php
```diff
@@ -589,15 +589,20 @@ private function parseResult($raw, $resultsType, $scrollDuration = null)
...
+                    '_index'  => $item['_index'],
+                    '_score' => $item['_score'],
+                ];
+                $output[$item['_id']] = array_merge($output[$item['_id']], $item['_source']);
```
saimaz: Can we avoid array_merge here?

einorler:
Removed checking if `_id` is returned, removed the returning of `index` and `type` in the result array and changed the implementation of results formation to avoid `array_merge` function

Service/Manager.php
```diff
@@ -590,14 +590,15 @@ private function parseResult($raw, $resultsType, $scrollDuration = null)
...
+            return [$data['_id'] => $data['_source']];
         }
 
         $output = [];
 
         if (isset($data['hits']['hits'][0]['_source'])) {
             foreach ($data['hits']['hits'] as $item) {
-                $output[] = $item['_source'];
+                $output[$item['_id']] = $item['_source'];
+                $output[$item['_id']]['_score'] = $item['_score'];
```
saimaz: `_score` is not needed at the array response

Service/Manager.php
```diff
@@ -590,14 +590,15 @@ private function parseResult($raw, $resultsType, $scrollDuration = null)
...
-            return $data['_source'];
+            return [$data['_id'] => $data['_source']];
         }
 
         $output = [];
 
         if (isset($data['hits']['hits'][0]['_source'])) {
             foreach ($data['hits']['hits'] as $item) {
-                $output[] = $item['_source'];
+                $output[$item['_id']] = $item['_source'];
```
saimaz: Do not change result structure. Append `_id` to the `$item` element.

einorler:
Removed `_source` from return array and changed back its structure, simply including an additional element with a key `_id`

PR #640, opened 2016-06-14
Implemented multi-search

### LINKED ISSUE #427: Support for multi-search
Feature request for https://www.elastic.co/guide/en/elasticsearch/reference/1.7/search-multi-search.html.

This should enable querying 50 queries at the same time in a single http request avoiding various overheads.

einorler:
closes #427


saimaz:
IMO implementation is a bit unuseful. What is the point add searches by expecting some time to execute? Much better would be to introduce new or extend the execute function by allowing to pass more than one search and use msearch endpoint for this.

PR #660, opened 2016-07-08
Implemented setting an id to persisted objects

### LINKED ISSUE #484: Set document ID after commit to ES

einorler:
PR should only be merged when the bundle has event dispatching support. The performance testing shows that the speed of indexing is decreased by 20 - 25 % compared to the implementation with event dispatching and ~40% when comparing to a current master branch implementation. My suggestion would be to continue increasing the performance and perhaps introduce a separate config value, that would allow to disable the id functionality when its not needed (during heavy indexing for example).

Closes #484

### DIFF
--- a/DependencyInjection/ONGRElasticsearchExtension.php
+++ b/DependencyInjection/ONGRElasticsearchExtension.php
@@ -48,6 +48,7 @@ class ONGRElasticsearchExtension extends Extension
                 $container->getParameter('kernel.debug') ? new Reference('es.tracer') : null,
             ]
         );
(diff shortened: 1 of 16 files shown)


saimaz:
Too slow.

saimaz:
Closing due slow performance. We need to find a better way or don't do at all.

Issue #428, opened 2015-09-25
Support for inner hits


Vesquen:
It would be pretty cool to have support for inner hits:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-inner-hits.html

saimaz:
This is the nice feature, but it is still experimental in the elasticsearch. I'm adding it to the v1.1 milestone and if that will become stable we will definitely add a support for it.

einorler:
At the moment it is possible to get raw inner hits from `Result::RESULTS_RAW` and `Result::RESULTS_RAW_ITERATOR`. Of course getting them isn't very user friendly that way, because one would have to go over a lot of levels of multidimensional arrays. Question is whether or not we want to introduce quite a lot of logic to the `DocumentIterator`, because the way I see it, a normal way to implement it would be to introduce logic similar to that of Aggregations.

toooni:
@saimaz @einorler The Problem (difference to Aggregations) I see here is that inner hits should be available within a document (or at least in the same result iteration somehow).
I am just starting to understand elasticsearch and this bundle - therefore I am sorry if I misunderstood something  ;)

I've thought about a few possible solutions:
- When the nested `_source` is excluded. It may make sense to hydrate the `inner_hits` directly into the collection?
- `AbstractDocument` class which implements properties to get and set (hydration) inner hits
- `InnerHitsDocumentTrait` which implements properties (must be used explicit if inner hits can happen on a `Document`

Only the first solution would break BC. The others can just extend functionality.

saimaz:
There are more and more inner hits support requests. Currently, I'm investigating how it can be iterated quickly as documents or something similar. 

For the quick inner hits access, there is possible to get a response from raw.

toooni:
I made a working solution where inner_hits are embedded into the document. I do have concerns that this doesn't match the requirements of @saimaz but wanted to give the input anyway.

I do know that this only makes sense as long as the childrens are excluded from `_source` because it would be overwritten by the hits anyway.
 
https://github.com/toooni/ElasticsearchBundle/commit/4310d2b0c89b846dcbf8272504d464506f30c9e8

saimaz:
Thanks for sharing your commit @toooni. I will borrow your commit and cherry pick to the new inner_hits iterator implementation.

toooni:
[...]

PR #663, opened 2016-07-15
[WIP] inner hit implementation in Document iterator

einorler:
This is the general idea that I have in mind. It's far from being finished or refined and I'm still not 100% sure about whether we need this at all.  Some feed back from the team would be much appreciated. @saimaz what are your thoughts about it?

Part of #428


saimaz:
It has to be implemented completely different. I'm closing.

PR #665, opened 2016-07-20
Disable conversion of 'null' date fields

martinemmert:
I recently had the problem that I worked on some legacy data sources where documents had date typed fields (_e.g. deletedDate_) set to `null`.

After I fetched such documents, these date fields were populated with unwanted `DateTime(0)` instances after converting the results into objects and lead to unexpected behavior or errors.

After adding these three lines the problems were gone and it worked as expected.

### DIFF
--- a/Result/Converter.php
+++ b/Result/Converter.php
@@ -91,6 +91,9 @@ class Converter
             if (isset($aliases[$name]['type'])) {
                 switch ($aliases[$name]['type']) {
                     case 'date':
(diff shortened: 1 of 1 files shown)


saimaz:
@martinemmert can you write a test case for this PR ?

martinemmert:
@saimaz I'll have to postpone this until we are finished - after that I will make myself some space to create the tests.

saimaz:
Solved in #770

PR #667, opened 2016-08-03
Made Manager::commit() return ids of persisted objects

### LINKED ISSUE #470: return documents id's in the commit() return
Would be usefull to get all isnerted elasticsearch document id's after commit. By default it returns `raw` response. What if we would add an option in the commit to return id's list.

einorler:
In addition I implemented throwing of an exception if there is an error in the `bulkResponse`. My only concern here is that currently a check is executed on persisted items to fetch the first occurring error message, but this involves an additional foreach loop. I would appreciate the opinion of others on this.

Closes #470

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -12,6 +12,7 @@
 namespace ONGR\ElasticsearchBundle\Service;
 
 use Elasticsearch\Client;
+use Elasticsearch\Common\Exceptions\ClientErrorResponseException;
 use Elasticsearch\Common\Exceptions\Missing404Exception;
 use ONGR\ElasticsearchBundle\Mapping\MetadataCollector;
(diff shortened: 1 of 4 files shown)


saimaz:
I cannot confirm that. 

I heard many times from inside people and from some outside NFQ that they would like to create a new project based on us but it stops that there is no checkout functionality. And nobody, absolutely nobody will use OXID or Magento or Shopware just for checkout. Some developer, for now, tries to use Sylius (http://sylius.org) components but most of them tell that is quite complex to modify.

My idea was to create a standard SDK (abstraction) for checkout workflow actions. Developers to create checkout solution will still need to finish the work in their project, but all standard actions will be there and it will accelerate process.

einorler:
Closing due to excess logic in `commit()`

PR #668, opened 2016-08-04
implemented throwing of exception in managers commit

### LINKED ISSUE #648: Throw exception on document commit error
Response from `$manager->commit()` is returned, but not checked for errors. Problems are hard to track, especially when new documents violate the mapping silently.

This is debatable, but generally, I vote in favour for throwing a custom exception by default. Cases are rare, when one would like to ignore such error and this could be easily configured using a parameter or try-catch block.

I strongly suggest, that exception must be thrown by default especially in debug environments.

einorler:
Closes #648

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -12,6 +12,7 @@
 namespace ONGR\ElasticsearchBundle\Service;
 
 use Elasticsearch\Client;
+use Elasticsearch\Common\Exceptions\ClientErrorResponseException;
 use Elasticsearch\Common\Exceptions\Missing404Exception;
 use ONGR\ElasticsearchBundle\Event\Events;
 use ONGR\ElasticsearchBundle\Event\BulkEvent;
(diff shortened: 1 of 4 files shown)


Service/Manager.php
```diff
@@ -12,6 +12,7 @@
 namespace ONGR\ElasticsearchBundle\Service;
 
 use Elasticsearch\Client;
+use Elasticsearch\Common\Exceptions\ClientErrorResponseException;
```
saimaz: Should be `use Elasticsearch\Exception\ClientErrorResponseException;`

Service/Manager.php
```diff
@@ -337,6 +340,12 @@ public function commit(array $params = [])
                     break;
             }
 
+            if ($bulkResponse['errors']) {
+                throw new ClientErrorResponseException(
```
saimaz: I think we could pass also the whole bulk response to the exception message. No?

einorler: Do you mean a json serialized bulk response array?

Service/Manager.php
einorler: No class with this namespace can be found

saimaz:
Please rebase PR

PR #673, opened 2016-08-23
Fix namespace in embedded multiple mapping example

alexander-schranz:


saimaz:
Thanks @alexander-schranz !

PR #678, opened 2016-09-01
Introduced strict result return type functions

saimaz:


Service/Repository.php
```diff
@@ -206,6 +208,82 @@ public function execute(Search $search, $resultsType = Result::RESULTS_OBJECT)
...
+
+
+    /**
+     * Returns DocumentIterator with composed Document objects from array response.
+     *
+     * @param Search $search
+     *
+     * @return DocumentIterator
+     */
+    public function findDocument(Search $search)
```
tautrimas: Is this a mistype? `findDocuments` should be plural, shouldn't it?

saimaz: Good point, it should be plural, otherwise, it means that only 1 document will be returned.

tautrimas: For documents it's pretty clear to understand what it means. But for arrays and raw below, I would suggest `findAsArray` and `findAsRaw`. Also `findAsDocuments` then. What do you think?

PR #682, opened 2016-09-12
Fix incorrect profiler query times

adrienbrault:
Hey, I noticed the time reported on the profiler toolbar/sidebar was different than the total of all queries.


saimaz:
Merged in #689

PR #702, opened 2016-11-09
Set index for bulk operations globally

### LINKED ISSUE #701: Manager->commit() not working when rest.action.multi.allow_explicit_index is set to false in elasticsearch config
For security reasons, on our ES server the option `rest.action.multi.allow_explicit_index` is set to `false` in our elasticsearch.yml.

When trying to call `$manager->commit()` I get the following error: "explicit index in bulk is not allowed".

This means that there must be no `_index`-setting in every single dataset in the body of the bulk query, but instead the index must be specified in the URL of the bulk endpoint. (See for instance https://github.com/elastic/elasticsearch-net/issues/870)

The solution to this problem would require the following changes in class `ONGR\ElasticsearchBundle\Service\Manager`:

- in method `bulk`, remove the setting `'_index' => $this->getIndexName()`
- in method `commit`, add `$bulkQueries['index']['_index'] = $this->getIndexName()` instead

wasinger:
Fixes #701

### DIFF
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -365,7 +365,7 @@ class Manager
     {
         if (!empty($this->bulkQueries)) {
             $bulkQueries = array_merge($this->bulkQueries, $this->bulkParams);
-
+            $bulkQueries['index']['_index'] = $this->getIndexName();
             $this->eventDispatcher->dispatch(
                 Events::PRE_COMMIT,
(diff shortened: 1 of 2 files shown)


saimaz:
#705 will be merged to the master after 1.2.5 release.

wasinger:
Thank you very much for merging and the 1.2.5 release!

PR #710, opened 2016-11-15
no-array-mapping option added to create index

carlcasbolt:
I've been having some issues with using array type data within my ES index.

When I try to create the index on my test machines on travis I get the following error:

```
  [Elasticsearch\Common\Exceptions\BadRequest400Exception]
  {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"No handler for type [array] declared on fie
  ld [qualifiers]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [statistic]: No handler f
  or type [array] declared on field [qualifiers]","caused_by":{"type":"mapper_parsing_exception","reason":"No hand
  ler for type [array] declared on field [qualifiers]"}},"status":400}
```

There is nothing wrong with the index as you can use array types, but if I need the array type defined in my model to work correctly for indexing and querying. However when attempting to create the index with the CLI tool I get hit by this error.

I might not be the best solution as `--no-array-mapping` on the `ongr:es:index:create` but I cannot see anywhere else in the code to correct this issue.

If there is a known work-around then I'm happy to adjust my code.

### DIFF
--- a/Command/IndexCreateCommand.php
+++ b/Command/IndexCreateCommand.php
@@ -39,7 +39,18 @@ class IndexCreateCommand extends AbstractManagerAwareCommand
                 InputOption::VALUE_NONE,
                 'If the time suffix is used, its nice to create an alias to the configured index name.'
             )
-            ->addOption('no-mapping', null, InputOption::VALUE_NONE, 'Do not include mapping')
+            ->addOption(
+                'no-mapping',
+                null,
+                InputOption::VALUE_NONE,
+                'Do not include mapping'
+            )
+            ->addOption(
+                'no-array-mappings',
+                null,
+                InputOption::VALUE_NONE,
+                'Do not include array mapping'
+            )
             ->addOption(
                 'if-not-exists',
                 null,
@@ -88,7 +99,7 @@ class IndexCreateCommand extends AbstractManagerAwareCommand
             return 0;
         }
 
-        $manager->createIndex($input->getOption('no-mapping'));
+        $manager->createIndex($input->getOption('no-mapping'), $input->getOption('no-array-mappings'));
 
         $io->text(
             sprintf(
--- a/Service/Manager.php
+++ b/Service/Manager.php
@@ -488,16 +488,34 @@ class Manager
      * Creates fresh elasticsearch index.
      *
      * @param bool $noMapping Determines if mapping should be included.
+     * @param bool $noArrayMappings Determines if array mapping should be included.
      *
      * @return array
      */
-    public function createIndex($noMapping = false)
+    public function createIndex($noMapping = false, $noArrayMappings = false)
     {
         if ($noMapping) {
             unset($this->indexSettings['body']['mappings']);
         }
-
+        if ($noArrayMappings) {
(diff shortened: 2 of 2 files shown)


einorler:
@carlcasbolt if I understand you correctly, then there is no need for any workarounds and you can index normal `string` or `integer` or any other property fields as arrays. There is no need for a separate `array` type of property annotation, so, if `qualifiers` field is a set of strings, your annotation should be:

```php
/**
 * @var array
 *
 * @ES\Property(type="string")
 */
```

If, however `qualifiers ` is a set of objects, use embedded or nested notation. Check out the documentation for further details:
http://docs.ongr.io/ElasticsearchBundle/mapping

carlcasbolt:
Hi @einorler 

I started off looking at Embedded objects because my array is a key value pair

e.g.
```
"qualifiers": {
    "minutes": "",
    "seconds": "",
    "angle": "",
    ...
}
```

However there are close to 300 qualifiers which vary from document to document, which is why I was looking at an array. Is there a way to create an Embedded object with the `__call()` method and a long list of params over 600 concrete methods?

Thanks.

saimaz:
There is possible to use public properties in the document, so in this case, you won't need getters and setters. But still, 300 are a lot. One case ould be to use dynamic templates to define your objects. The other way would be to use different structure for the object, e.g. like:

```json
{
"qualifiers": 
   [
     {
        "type": "minutes",
        "value": "60"
     },
     {
        "type": "seconds",
        "value": "12"
     }
  ]
}
```

carlcasbolt:
Hi @saimaz

I considered that way of storing the data as well but it did not allow me to generate meaningful aggregations on the data as I'm using ES to generate stats on millions of documents.

saimaz:
@carlcasbolt can you tell me what results you want to get?

carlcasbolt:
I will be storing a few million records in ES to generate stats.

I am have a type set on each document and then qualifiers which apply to a given type.

I will be generating aggregations on those types to show counts and groups of this data.

E.g. a pass, which has qualifiers such as minute, second, x, y, angle, distance, speed.

From this I will generate aggregations of the average distance of the passes, the average speed, the count of them within a given radians of angle to show forward, backwards and sideways. The x & y co-ords will be used to provide stats on the number of passes from a given region to another. 

In the instance of a goal I will have different qualifiers, which would be x, y and z to allow of 3d space within a goal so that I can get elevation as well as location within the area. I have the originating location of the shot on goal as a separate set of of x and y. I then have the same minute and second qualifiers to allow me to give stats on the average time of a goal or period of a game.

In the instance of a foul I have qualifiers such as type of foul, if there is a body part involved, the outcome of the foul, was a card, penalty, free-kick awarded. Again we have x and y co-ords to generate stats on the areas of the pitch where fouls happen most often.

The list goes on, but hopefully you can get a feel on what I'm using the aggregations for

saimaz:
Arrays support now will be available in bundle's v5 through hashmaps annotation.

carlcasbolt:
Thanks @saimaz - can you point me in the direction of the documentation?

Issue #620, opened 2016-04-21
Index drop in IndexImportCommandTest.php


einorler:
There is an irrational drop of index in both tests in IndexImportCommandTest.php, after that, the tests attempts to execute a search to the freshly dropped index:
https://github.com/ongr-io/ElasticsearchBundle/blob/b305614610c264e4da6b3097b536c383e15d6add/Tests/Functional/Command/IndexImportCommandTest.php#L76

This results in 7 tests 0 assertions. What is even more alarming is that after trying to comment these lines out, the tests fail and the apparent reason is that the data is not being loaded to the ES.

saimaz:
Fixed in 5.0 rc

PR #713, opened 2016-11-16
Fixing abstract elastic search test exception throwing.

saimaz:


einorler:
closes #620

einorler:
I got rid of the majority of errors that appeared after removing the elasticsearch exception ignore. The one left is in the test here:
https://github.com/ongr-io/ElasticsearchBundle/blob/09a07f00e5930a9d5e07e7f7de1f38c54365b63a/Tests/Functional/Result/DocumentScanIteratorTest.php#L90

It seems that with data set `#1` it is unable to clear the scroll in manager after the test (called in the destructor of document iterator). ES client throws an error with no message. I don't really know how to handle it. Any help would be appreciated.

saimaz:
Document scan iterator is deprecated and no longer exists. There is no reason to have this test case. Better just cover what is more important.

PR #717, opened 2016-11-17
rename "annotations.cached_reader " service.

bangpound:
I am seeing very peculiar things happen on my project. All of the annotation cache -- Serializer, Router, Doctrine -- is being written to the cache directory for this bundle.

The service definition has the same name as a private service in Symfony framework. Can you rename it?


saimaz:
Thanks for the notice. Absolutely I can rename it in the master for `2.0`. I cannot rename it in `1.2.x` branch because it's a breaking change.

Issue #725, opened 2016-12-27
Can not fetch cache $mapping


weburnit:
`
vendor/ongr/elasticsearch-bundle/Mapping/MetadataCollector.php:113
`
```php

        $this->enableCache && $mappings = $this->cache->fetch($cacheName);

        if (isset($mappings) && $mappings) {
            return $mappings;
        }
```
Instead of call function contain(), we must use fetch() to retrieve correct $mappings.

Bundle version v5.0

weburnit:
on `prod` env, we can not have mapping!

weburnit:
@saimaz I've created 2 pull request which fixed my application. Would you plz approve it?

weburnit:
Fixed

PR #728, opened 2016-12-28
Fix If clause which retrieve incorrect data from cache

weburnit:
Similar to #725 
We have several place in this class which have defect condition through IF_clause


saimaz:
Please open PR to `5.0` branch.

weburnit:
I guess we must add more UnitTest!
It looks quite simple!

saimaz:
Sure we need more unit tests. I'd appreciate your future contributions @weburnit.

PR #730, opened 2016-12-28
Fix If clause which retrieve incorrect data from cache (#728)

weburnit:


saimaz:
Are you using `dev-master`? Please consider using 5.0-dev instead. 
See the http://docs.ongr.io/common/Contributing 
After fix or future merged to particular version branch, I will handle merging upper, in this case to the master.  

I'm closing this PR.

PR #732, opened 2017-01-03
Fix cache checking

norkunas:


einorler:
Why did you duplicate the condition?

norkunas:
Accidentally :/

weburnit:
2previous PR has fixed this one @norkunas !

norkunas:
@weburnit this is the different method i've updated

norkunas:
https://github.com/ongr-io/ElasticsearchBundle/blob/master/Mapping/MetadataCollector.php#L344 so the branches are different?

PR #749, opened 2017-02-07
_id is optional at json dump

asev:
In documentation it says, that `_id` is optional, but it has been marked as required field.

> `_id` is optional, if not specified (provided `null`) elasticsearch will create a random id for that document.

https://github.com/ongr-io/ElasticsearchBundle/blame/master/Resources/doc/commands.md#L66


saimaz:
Hi @asev, nice catch. Can you do this change from `1.2` branch? In that case I could release a fix patch from 1.x version.

PR #755, opened 2017-02-21
load Embedded metaFields

juliensantos87:
if Embedded metaFields are not load on the first pass
cached metaData will be wrong (without _id for exemple)


saimaz:
Thank you @juliensantos87 for the PR.

jazby:
I'm not confused.
I am happy that the patch is ready. Just sad, because the bug was not immediately find and determine that was fixed "a while ago" is slightly frustrating :)
Thank you for the correction, and I hope that will soon be in a version, it's blocking me.
By the way, what progress the pull 670? I would very much needed it.

PR #764, opened 2017-03-16
Missing metaFields on embedded property. 

jazby:
I have 2 nested entities (SubA, SubB) in one document (MasterA). Metadata is generated in this order: SubA, MasterA, SubB.

SubA is generated using method 'parse', SubB is generated using method 'getAliases'. In the second case the variable $metaFields is missing when the recursion calls getAliases.

So - if the metadata is generated, the first variable is skipped in the nested object (SubB) because of this error. This only happens if the es.cache is set to true.

The result is that if you have more nested entities in one main entity and caching is enabled, the first parameter is present only in the first nested object and in all other following nested objects is missing.

Mysterious bug.

### DIFF
--- a/Mapping/DocumentParser.php
+++ b/Mapping/DocumentParser.php
@@ -283,7 +283,7 @@ class DocumentParser
                         [
                             'type' => $this->getObjectMapping($type->class)['type'],
                             'multiple' => $type->multiple,
-                            'aliases' => $this->getAliases($child),
+                            'aliases' => $this->getAliases($child, $metaFields),
                             'namespace' => $child->getName(),
                         ]
                     );


saimaz:
It's already fixed in #755 and on this way to the master branch.

saimaz:
I see you are confused. The explanation is quite simple. When something is fixed in earlier version it will be merged up to the master branch. So if you want to have this fix, you can use `dev-master` or 5.x dev branch. I already merged it from 1.2 branch. And soon will release a patch with included fix for 1.2.x bundle version.

jazby:
I'm not confused.
I am happy that the patch is ready. Just sad, because the bug was not immediately find and determine that was fixed "a while ago" is slightly frustrating :)
Thank you for the correction, and I hope that will soon be in a version, it's blocking me.
By the way, what progress the pull 670? I would very much needed it.

saimaz:
Released fix in https://github.com/ongr-io/ElasticsearchBundle/releases/tag/v1.2.8

Issue #763, opened 2017-03-16
Mapping parsing fails when using embedded objects with custom doc dir


einorler:
In 5.0 when using custom directory like so:

```yml
mappings:
  AppBundle:
    document_dir: CustomDirectory
```

The mappings fail to be parsed if embedded or nested objects are used. If, for example, I were to use an embedded `Variant` object in the document `Product` that are in the `Entity` namespace like so:

```php
namespace AppBundle\Entity;

use ONGR\ElasticsearchBundle\Annotation as ES;

/**
 *
 * @ES\Document()
 */
class Product
{
 // ...

    /**
     * @var Variant[]
     *
     * @ES\Embedded(class="TestBundle:Variant", multiple=true)
     */
    public $variants;

  // ...
}
```

I would get an error that would say that the class `AppBundle\Document\Variant` could not be found.

einorler:
Related issue: #729 :

https://github.com/ongr-io/ElasticsearchBundle/blob/12f44a212dbec07146f8bd5f8465c5599669a14b/Mapping/DocumentParser.php#L554

einorler:
Update: It works if I use the full namespace instead of `Bundle:Class` in embedded annotation

PR #765, opened 2017-03-17
Separate dir mapping fix

einorler:
This is a potential temporary fix for the problem at hand. It is completely backwards compatible but it only reduces the risk of such an issue, but does not neglect it completely. The problem here is that every manager can potentially have different document directories and the metadata collector and document parser that handle the parsing of annotations are not concerned at all with what manager is being parsed at any given moment. This means that in order to solve this issue the right way, we would have to change the entire concept of how metadata is being handled. 

Either that or I missed something out, if so, let me know. @saimaz what do you think?


einorler:
Closes #763

PR #769, opened 2017-03-23
[BUGFIX] Avoid php warnings when type="object" is used

PortaltechGithub:
I'am using plain type="object" in my mapping annotation to index simple array structures. I don't need some kind of specified nestet object here, so the only solution is do set type to "object".

Without the fix i got warnings on console. In my dev env and test env warnings are handled as errors, so my app breaks and indexing stops.

This little fix checks if namespace and aliases from reflection stuff is set.

### DIFF
--- a/Result/Converter.php
+++ b/Result/Converter.php
@@ -110,11 +110,13 @@ class Converter
                             if (!isset($value)) {
                                 break;
                             }
-                            $value = $this->assignArrayToObject(
-                                $value,
-                                new $aliases[$name]['namespace'](),
-                                $aliases[$name]['aliases']
-                            );
+                            if (isset($aliases[$name]['namespace']) && isset($aliases[$name]['aliases'])) {
+                                $value = $this->assignArrayToObject(
+                                    $value,
+                                    new $aliases[$name]['namespace'](),
+                                    $aliases[$name]['aliases']
+                                );
+                            }
                         }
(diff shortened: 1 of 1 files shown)


einorler:
Could you be more precise with what kind of arrays you need, maybe give an example?

einorler:
If you just need an array like this `"types": ["type_1", "type_2", "type_3"]`, then you shouldn't use object notation, just simple keyword or text notation will work fine.

saimaz:
@PortaltechGithub you also can use hashmap annotation to have a more dynamic support for object like fields.

saimaz:
I'm closing this issue because object type in the `ElasticsearchBundle` is used only with the embedded fields.

PortaltechGithub:
Sry, i did not get email notifications of your comments.

In my use case i have a multi dimensional array and just want to store this array in an object. To use a field type keyword doesn't work in my case.

Example:
```
$arr = [
   'foo' => [
      'bar' => 'some text',
      'foobar' => ['data' => 'some more text']
   ]
]
```
This data is stored as json object by elasticsearch. The problem with NestedObjects is they are not dynamic enough. I have to declare every single property and in my case i only need those data for diplaying it in results.

This kind of mapping is documented here: https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html

PR #777, opened 2017-04-04
Make ObjectIterator an AbstractLazyCollection

### LINKED ISSUE #774: Support Common Collections?

mcrumm:
In #775 I was able to implement the custom hydration logic into most methods of `ArrayCollection`.  However, numerous methods operate directly on the internal `$elements` object, so keeping that in sync with the `$rawObjects` list becomes cumbersome. Any performance improvement gained from hydrating on-demand is lost as soon as a user wants to run a `map()` or a `filter()` over the collection.

This PR, which could be used in place of #775, implements `ObjectIterator` as an `AbstractLazyCollection`, which will hydrate all the raw data, but only once the data has been requested.  If a document contains a collection, but no operations are ever performed on that collection, no hydration will take place.

Ultimately, this ensures that we avoid any potential side effects associated with performing operations on a document's collection.  It comes at the cost of hydrating all items in the collection once, but for the most common use cases, that cost will likely be incurred during iteration anyway.

Closes #774

### DIFF
--- a/Collection/Collection.php
+++ b/Collection/Collection.php
@@ -11,107 +11,13 @@
 
 namespace ONGR\ElasticsearchBundle\Collection;
 
+use Doctrine\Common\Collections\ArrayCollection;
+
 /**
  * This class is a holder for collection of objects.
+ *
+ * @deprecated Use Doctrine\Common\Collections\ArrayCollection instead.
  */
-class Collection implements \Countable, \Iterator, \ArrayAccess
+class Collection extends ArrayCollection
 {
-    /**
-     * @var array
-     */
-    private $elements = [];
-
-    /**
-     * Constructor.
-     *
-     * @param array $elements
-     */
-    public function __construct(array $elements = [])
-    {
-        $this->elements = $elements;
-    }
-
-    /**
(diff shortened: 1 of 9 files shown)


mcrumm:
@saimaz The single failed test job appears to be a timeout, rather than an actual problem.  You may want to try re-running the Travis build.

mcrumm:
@saimaz Any thoughts on this implementation?

saimaz:
So far I think it's a great PR. Since it's a big change, I want to test it and play around with it entirely, to make sure everything is good.

PR #780, opened 2017-04-10
Update elasticsearch index if exists

kmiladi:


kmiladi:
must add tests

saimaz:
Hi @kmiladi sorry I didn't respond quickly. I'm not sure it's a good practice to update mapping after index is created. The creators of Elasticsearch recommends to create a new index instead. 

Here's the blog post about it: https://www.elastic.co/blog/changing-mapping-with-zero-downtime

kmiladi:
Hello @saimaz , no problem
Ah ok if they recommended this, so we must floow :+1:

saimaz:
It totally makes sense since the type changes cause very often conflicts. And the only way to solve it is to create a new index. 

It raises a new topic how to reindex the data quickly, and maybe automate this process, like creating the new index, reindex the data and only then switch alias. I'm still thinking about the possible implementation of those features if you have any ideas you are welcome to suggest ;) or open a PR for that.

kmiladi:
Really, I don't want to modify the mapping of the existing index, I want to add other mappings to the existing index. 

I don't know if it changes anything for you?

Thanks

saimaz:
When you create the index it put's mapping already. There is an option to skip mapping in very specific cases if you need only.

kmiladi:
Yes but in the php extension elasticsearch, we have a methode putMapping to put a new mapping in the index

kmiladi:
public function putMapping($params)
    {
        $index = $this->extractArgument($params, 'index');

        $type = $this->extractArgument($params, 'type');

        $body = $this->extractArgument($params, 'body');

        /** @var callback $endpointBuilder */
        $endpointBuilder = $this->endpoints;

        /** @var \Elasticsearch\Endpoints\Indices\Mapping\Put $endpoint */
        $endpoint = $endpointBuilder('Indices\Mapping\Put');
        $endpoint->setIndex($index)
                 ->setType($type)
                 ->setBody($body);
        $endpoint->setParams($params);

        return $this->performRequest($endpoint);
    }

saimaz:
I believe I misunderstood something. What you are showing is native client. This is a bundle, which uses it. To create the index you have to define a manager with all information and documents. Then bundle reads documents annotations and create a mapping config and passes it to create method when index is created.

Here's the exact place:
https://github.com/ongr-io/ElasticsearchBundle/blob/162bbe17538c9dfec2e6793eaca96b40976aeba1/Service/Manager.php#L505

kmiladi:
Yes I see, but what is the utility of the putMapping method (even if it is not part of your bundle)? Logically it's for put a new mapping in the index, what do you think?

saimaz:
In that particular case you can get a client instance from this bundle manager and call any function you like.

something like this:
```
$manager = $this->get('es.manager.default');
$mapping = $manager->getIndexMappings();
$client = $manager->getClient();
$client->putMapping(
    [
      'index' => $manager->getIndexName(),
      'mapping' => $mapping
.... //Not sure currently what is the syntax of native client, just guessing :)
    ]
);
```

kmiladi:
That's exactly what I did
    private function updateIndex($manager)
    {
        foreach ($manager->getIndexMappings() as $key => $mapping) {
            $params = [
                'index' => $manager->getIndexName(),
                'type' => $key,
                'body' => $mapping
            ];

            $manager->getClient()->indices()->putMapping($params);
        }
    }

kmiladi:
public function updateIndex($manager)
    {
        foreach ($manager->getIndexMappings() as $key => $mapping) {
            $params = [
                'index' => $manager->getIndexName(),
                'type' => $key,
                'body' => $mapping
            ];

            $manager->getClient()->indices()->putMapping($params);
        }
    }

saimaz:
Yes, what I'm trying to say is if you need this you should do it in your project. Create separate service or smth. As Elasticsearch recommends, create new index instead of change mapping due to some complications. I don't want to add such functionality to the manager.

PR #781, opened 2017-04-11
Generate is Methods for boolean properties.

fattouchsquall:

PR #782, opened 2017-04-11
Add isser for boolean properties.

fattouchsquall:
Hello @saimaz,

We can close the other PR https://github.com/ongr-io/ElasticsearchBundle/pull/781, i created a new one from branch 5.0.


saimaz:
Cool, thanks for the PR.

fattouchsquall:
Hello @saimaz,

Can you check plz my fix for the case when using document_dir: Document\SubDocument,

Thanks

saimaz:
Good point. Would be really great to have a test case for this as well.

fattouchsquall:
i'm on it

fattouchsquall:
@saimaz,

I've added tests for this fix. Can you check them plz?

saimaz:
There are already 2 documents in the custom fixtures dir, you can take as an example: `Tests/app/fixture/TestBundle/Entity/Product.php`. In that case, we don't need to introduce a new one.

fattouchsquall:
Done @saimaz

PR #784, opened 2017-04-13
Add visibility property when generating a document.

fattouchsquall:
This PR adds a visibility for a generated document's field (protected or private) for fixing inheritence from document class.
Also it fixes the default type text for elastic in generation of document.


saimaz:
Great PR. The only thing I would add here is to skip getters and setters generation if the type is public.

fattouchsquall:
That's why i handle only protected and private visibilities and not public, but i can add it, i agree.

fattouchsquall:
@saimaz ,

I add the visibility public and remove generation of getter and setter when the visibility is public. can you check it plz.

Issue #793, opened 2017-06-19
Exception: [trait] class cannot be parsed as document because @Document annotation is missing


kategray:
I am using PHP 7.0 with the elastic search bundle.  Document annotations are being included using a trait to avoid code duplication.

There is an exception that is being caught,  '"DirectoryBundle\Document\BusinessTrait" class cannot be parsed as document because @Document annotation is missing.' on page load.

DocumentFinder is looking for all PHP files.  This includes traits.

```$files = new \RegexIterator($iterator, '/^.+\.php$/i', \RecursiveRegexIterator::GET_MATCH);```

Traits aren't really exceptions - rather than catching them, it makes more sense to use the reflection classes ability to indicate if the file contains a trait.

PR #794, opened 2017-06-19
Don't throw an exception when a Trait is found in the same directory

kategray:
Patch for issue #793


saimaz:
Nice PR! Thank you @kategray.

PR #804, opened 2017-07-25
Fixed embed include for extended documents

### LINKED ISSUE #803: Mapping for Embed not working when its extended
- PHP: 7.0
    - ElasticSearch: 5
    - Ongr: 5.0.2

It use the false folder when an embed is from an extended Document: `https://github.com/ongr-io/ElasticsearchBundle/blob/v5.0.2/Mapping/DocumentParser.php#L599`. The `guessDirName` returns for `SuluArticleBundle:ExcerptViewObject` the `Bundle/WebsiteBundle/Document/` instead of `Document`. A strange namespace is then generate `Sulu/Bundle/ArticleBundle/Bundle/WebsiteBundle/Document` in https://github.com/ongr-io/ElasticsearchBundle/blob/v5.0.2/Mapping/DocumentParser.php#L572.

Folder Structure:

```
 - src
   - Client
      - Bundle
        - WebsiteBundle
          - Document
            - `ArticleViewDocument` extends ArticleBundle:ArticleViewDocument

 - vendor
    - sulu
       - article-bundle
         - Document
           - `ArticleViewDocument`
           - `ExcerptViewObject`
```

ArticleBundle:ArticleViewDocument

```php
    /**
     * @var ExcerptViewObject
     *
     * @Embedded(class="SuluArticleBundle:ExcerptViewObject")
     */
    protected $excerpt;
```

alexander-schranz:
The document parser should never care about how to find a document. If you use the same directory as in it will not work when you Extend your Document. This did work correctly in 1.* version.

fixes #803

### DIFF
--- a/Mapping/DocumentParser.php
+++ b/Mapping/DocumentParser.php
@@ -197,11 +197,10 @@ class DocumentParser
      * Returns meta field annotation data from reader.
      *
      * @param \ReflectionProperty $property
-     * @param string              $directory The name of the Document directory in the bundle
      *
      * @return array
      */
-    private function getMetaFieldAnnotationData($property, $directory)
+    private function getMetaFieldAnnotationData($property)
     {
         /** @var MetaField $annotation */
         $annotation = $this->reader->getPropertyAnnotation($property, self::ID_ANNOTATION);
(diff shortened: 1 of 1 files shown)


saimaz:
This cannot pass the mapping test case. I will take a look later what's the problem here, but you can help with this as well, and I will merge ;)

alexander-schranz:
@saimaz seems like the custom directory is not longer working with this change. If I understand the custom directory correctly it needs to get the custom directory from the Manager. As the is no Manager available in the Converter, MetadataCollector or DocumentParser the custom directory need to be grinding throw all functions from manager to the parser this way. But maybe the change in #817 should be enough.

alexander-schranz:
close in favor of #817

PR #816, opened 2017-10-10
results converter continue fix

jazby:
In method assignArrayToObject should be continue 2, otherwise it will end the iteration of the rest values.


saimaz:
Thank you for the PR @jazby. Is there any chance to cover this with a test example?

jazby:
Sorry for mistake. You'r version is correct. Closed.

PR #817, opened 2017-10-12
Mapping now based on property instead of class

### LINKED ISSUE #803: Mapping for Embed not working when its extended

kleinkoerkamp:
closes #803.

Because of inheritance the mapping was always based on the highest level class, now looking to the class where the property is defined in.

### DIFF
--- a/Mapping/DocumentParser.php
+++ b/Mapping/DocumentParser.php
@@ -250,7 +250,6 @@ class DocumentParser
     private function getAliases(\ReflectionClass $reflectionClass, array &$metaFields = null)
     {
         $reflectionName = $reflectionClass->getName();
-        $directory = $this->guessDirName($reflectionClass);
 
         // We skip cache in case $metaFields is given. This should not affect performance
         // because for each document this method is called only once. For objects it might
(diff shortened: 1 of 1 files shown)


alexander-schranz:
if this fixes #803 this would make #804 unnecessary for me.

alexander-schranz:
Any updates to this @saimaz? /cc @kleinkoerkamp

PR #826, opened 2017-11-16
Fix usage of fields as reference

### LINKED ISSUE #772: Parent Mapping does not work
I tried to map a parent document via annotation. I debugged some code and can't find the spot where `_parrent` is mapped to index. My configuration from the property is stored in aliases, but then the journey ends :)

In `\ONGR\ElasticsearchBundle\Mapping\DocumentParser::getAliases()`  calls `$this->getMetaFieldAnnotationData($property)`. The correct parent type appears in `settings`, but this informations is not further used.

Also in \ONGR\ElasticsearchBundle\Mapping\MetadataCollector::getClientMapping the config will be cropped from the $mapping.

I think this function is totaly broken, but i can't find the right spot to provide a bugfix.

toooni:
90c505a introduced a bug where aliases were not merged in fields.
The problem here is that the second parameter of getAliases is submitted by reference with the intention of using it in fields.
I do recommend solving this a bit different because using parameters by reference is often a bad idea and can lead to problems like this one.

However this reverts the error introduced in 90c505a and fixes #772

### DIFF
--- a/Mapping/DocumentParser.php
+++ b/Mapping/DocumentParser.php
@@ -111,7 +111,7 @@ class DocumentParser
             }
 
             $fields = [];
-            $aliases = $this->getAliases($class, $fields)
+            $aliases = $this->getAliases($class, $fields);
 
             $this->documents[$className] = [
                 'type' => $document->type ?: Caser::snake($class->getShortName()),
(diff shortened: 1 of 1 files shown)


saimaz:
For some reason, Travis is angry on $this usage in your PR. I will double check and if everything ok will release this in the next patch.

toooni:
@saimaz Sorry, I missed a semicolon in the hurry.

saimaz:
thanks for the PR @toooni

PR #829, opened 2017-12-28
fixed twig template references for symfony 3.4

nu-fp:
When I try to install the ElasticsearchBundle with symfony 3.4 I get an profiler error on the bottom of the site, stating "An error occurred while loading the web debug toolbar. Open the web profiler.". Opening the web profiler reveals an error: "The profiler template "ONGRElasticsearchBundle:Profiler:profiler.html.twig" for data collector "ongr.profiler" does not exist."

I found two files, where the references to twig templates crashes as of symfony 3.4, due to the new Twig namespace and changed the code:
* Resources/config/services.yml
* Resources/views/Profiler/profiler.html.twig
These have to be changed, in order to work with symfony 3.4 and above (https://symfony.com/doc/3.4/templating.html#referencing-templates-in-a-bundle).

I tested this code with symfony 3.1 & 3.4.

### DIFF
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -1,6 +1,5 @@
 parameters:
     es.logging.path: "%kernel.logs_dir%/elasticsearch_%kernel.environment%.log"
-    es.profiler.template: ONGRElasticsearchBundle:Profiler:profiler.html.twig
 
 services:
     es.export:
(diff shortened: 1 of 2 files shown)


saimaz:
Thank you for the PR @nu-fp

PR #835, opened 2018-02-15
explicitly register commands as services, so that support Symfony 4

tiger-seo:


.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea/
```
saimaz: can you remove .idea in this file?

tiger-seo:
@saimaz done

saimaz:
Thanks for the PR @tiger-seo

trandangtri:
Hi @saimaz,
With this PR, we won't support Symfony 2.8 and older, is it right?

CC @tiger-seo

saimaz:
yes, by releasing a 5.1 we can drop support for the Symfony 2.8

PR #840, opened 2018-05-03
6.0 major version upgrade

saimaz:


saimaz:
Here's some sneak peak for the configuration:

```
ongr_elasticsearch:
    _defaults:
        number_of_shards: 4
        refresh_interval: -1
        number_of_replicas: 0
        hosts:
            - 127.0.0.1:9200
    analysis:
        filter:
            incremental_filter: #-> analyzer name
                type: edge_ngram
                min_gram: 1
                max_gram: 20
        analyzer:
            incrementalAnalyzer: #-> analyzer name
                type: custom
                tokenizer: standard
                filter:
                    - lowercase
                    - incremental_filter
    indexes:
        ongr-product-test:
            settings:
                number_of_shards: 2 #-> default 4, and override with 2
            mapping: TestBundle\Document\Product
```

saimaz:
I have some idea also to make configuration as less as possible. In that case, maybe would make sense to set an index name in the `@Document` annotation. There would be additional options like: `index` and `index_settings`.

saimaz:
It's a prototype. With the 6th version and further development, I don't see the benefits of keeping bc with the 5th version and sf 3. There are so many changes in the symfony and elastic. The only downside is in the project you have to upgrade it. During the time and using this bundle I learned a lot and found some logic mistakes which I want to fix. Also, the intention for these changes is related to two other very useful bundle development and new major version support, the FilterManagerBundle and RouterBundle.

Service/Manager.php
```diff
@@ -1,727 +0,0 @@
...
-use ONGR\ElasticsearchBundle\Exception\BulkWithErrorsException;
-use ONGR\ElasticsearchBundle\Mapping\MetadataCollector;
-use ONGR\ElasticsearchBundle\Result\Converter;
-use Symfony\Component\EventDispatcher\EventDispatcherInterface;
-use Symfony\Component\Stopwatch\Stopwatch;
-
-/**
- * Manager class.
- */
-class Manager
```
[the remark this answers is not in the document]
saimaz: @alexander-schranz I wrote a comment in the PR regarding this and all the changes.

alexander-schranz:
@saimaz 

> Maybe would make sense to set an index name in the @Document annotation

As index name likes database name mostly come from the parameters.yml or env variable I think this is not a good idea to move the index_name to the Document Annotation. At least a index_prefix_name should be configureable in the configuration. 

> The only downside is in the project you have to upgrade it

For a project its always easy to upgrade as you can make there breaking changes but for a reusable bundle its not.

alexander-schranz:
@saimaz If we don't have a manager how do we choose where to index a document if you look at our [current configuration](https://github.com/sulu/SuluArticleBundle/blob/1.0.0/Resources/doc/installation-es5.md) we are indexing the same document into 2 indexes (based on some business logic its not always indexed in live index). How I could achieve this without having manager as normally you get the correct repository from the manager.

saimaz:
The thing is that from the 6th version there is no need to have a manager anymore. The manager was to support the index with more than one type. From now on every type is an index. I don't want to introduce BC changes so will keep the manager with the existing functions, but will mark it as deprecated and will remove later on in 7th or 8th version. 

For the configuration, I imagine that BC can be kept as well. With some additional checks should be everything fine. But later on also will be removed and everything would be great to keep as clean as possible. This will let to work more easily with Symfony auto wiring and other new features.

alexander-schranz:
How you can index the same document in 2 different indexes based on business logic then?

Example configuration:

```
    indexes:
        admin-index:
            settings:
                number_of_shards: 2 #-> default 4, and override with 2
            mapping: TestBundle\Document\Page
        website-index:
            settings:
                number_of_shards: 2 #-> default 4, and override with 2
            mapping: TestBundle\Document\Page
```

So there need to be a index service or a paramter on the repository in which index the document should be stored?

saimaz:
The repository would be index representation. So simply you persist to the specific repository which works only one index.

chirimoya:
@saimaz can you give us a short update on the state of this PR? thx

saimaz:
This PR will be closed in favor of #868. It was my experimental version of what it could be and to ask community suggestions. 

You can check the [6.0-dev](https://github.com/ongr-io/ElasticsearchBundle/tree/6.0-dev) branch and functional tests on how the usage has been changed. There are not so many changes but rather complexity reduces. Full documentation, change list and bc changes, upgrade info are coming :) I didn't write it in a first place since I want to fully test the prototype which I released as an early [6.0.0-beta](https://github.com/ongr-io/ElasticsearchBundle/releases/tag/v6.0.0-beta).

DependencyInjection/Compiler/MappingPass.php
```diff
@@ -11,85 +11,101 @@
...
+        $container->setParameter(Configuration::ONGR_INDEXES, $indexes);
 
-            // Building repository services.
-            foreach ($mappings as $repositoryType => $repositoryDetails) {
-                $repositoryDefinition = new Definition(
-                    'ONGR\ElasticsearchBundle\Service\Repository',
-                    [$repositoryDetails['namespace']]
-                );
-                $repositoryDefinition->setPublic(true);
+        $defaultIndex = $defaultIndex ?? array_shift(array_keys($indexes));
```
toooni: If no index is present (f.e. in a new environment) `array_shift(array_keys($indexes))` is invalid.

Mapping/DocumentParser.php
```diff
@@ -13,434 +13,250 @@
...
-     *
-     * @param \ReflectionProperty $property
-     *
-     * @return Embedded|object|null
-     */
-    private function getEmbeddedAnnotationData(\ReflectionProperty $property)
+    public function getIndexMetadata($namespace): array
     {
-        $result = $this->reader->getPropertyAnnotation($property, self::EMBEDDED_ANNOTATION);
+        $class = new \ReflectionClass($namespace);
```
toooni: We should catch `ReflectionException` here. Filenames / Classes do not always match.
```
        try {
            $class = new \ReflectionClass($namespace);
        } catch (\ReflectionException $e) {
            return [];
        }
```

saimaz: Well if filenames and classes (namespaces) won't match it will break PSR-4, and composer will throw an exception. Anyway, thanks for the insight, will think about how to improve this place.

DependencyInjection/Compiler/MappingPass.php
saimaz: That is completely true. Thanks for the notice.

saimaz:
I'm closing this branch in favor of #868 and [v6.0.0-beta](https://github.com/ongr-io/ElasticsearchBundle/releases/tag/v6.0.0-beta) release. Will open a pull request to the master from 6.0-dev where currently is the latest state of a 6th major version upgrade. If you anyone noticed something wrong or have any idea to improve any place, please comment it out there.

Mapping/DocumentParser.php
toooni: No. You are iterating over the users project dir. You don't know whats hidden in there ;)
In my case it was a wrong defined test namespace. This isn't a static analysis tool.

saimaz: Ok, I misunderstood for the first time. The idea behind this is to iterate through the src directory and collect the files which have `@Index` annotation which indicates that is an ElasticSearch index. THen it grabs the namespace from that file and loads the reflection class. I agree that the exception has to be thrown if it cannot load it.

toooni: I saw that, yes. I wonder if there's another way which might be a bit more performant?
Do you think it might be too complicated for the users to just allow a specific folder like `Document`?

saimaz: I was thinking about a more performant way to get the mapping. But it will be executed once on warmup and then cached. So not a big deal if it will take 1-second longer IMO. 

Regarding the additional categories, I already have introduced a specific parameter for that.

Take a look at the #869 It's the latest state there and potential RC version which most likely won't change drastically.

toooni: Sorry the `Document` folder think was related to the performance. I meant to require all Documents to be inside a `Document` folder. This way not all classes are checked. But as long as it's really just a second - I think you're right.

PR #845, opened 2018-07-09
rename Object class to ObjectType, fix mentions of it

alulord:
fix Object class name, for php 7 support


saimaz:
There are some problems with the Symfony 2.8, but I think for PHP 7.2 support we can release 5.1 version by dropping old Symfony 2.8 version.

alulord:
There are, but as you can see [here](https://travis-ci.org/ongr-io/ElasticsearchBundle/builds/373902894?utm_source=github_status&utm_medium=notification)  those issues are already in master branch. I think I did identify the problem in this config https://github.com/ongr-io/ElasticsearchBundle/blob/master/Resources/config/services.yml
specifically this notation:
```
services:
    ONGR\ElasticsearchBundle\Command\:
        resource: '../../Command'
        tags:
            - { name: console.command }
```
which I believe is not support by symfony 2.8. I can fix it in this PR, howevere since it is in master already a new PR would be more feasible (also I don't have time to do it now)

PR #852, opened 2018-10-19
PHP7.2 compatibility. BC compatible.

makasim:


saimaz:
Looks quite good, but I'm away and only have phone with me. Will review in more detail on Monday and merge then ;) Thank you for the PR.

saimaz:
Renaming `@ObjectType` will create BC break but there is no other choice. Also renaming only object with the type we are breaking the consistency. IMO the Nested also should be with the type suffix and maybe others as well. This is not very important but in the ElasticsearchBundle v6 I will change completely it.

PR #858, opened 2018-11-15
Restrict compatible PHP versions to currently supported ones

YetiCGN:
We haven't tested PHP 7.3, 7.4, 8.0 or 9.0 yet, so we shouldn't claim compatibility with *every* version greater than or equal to PHP 5.6.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": ">=5.6",
+        "php": ">=5.6,<7.3",
         "symfony/framework-bundle": "^2.8|^3.0|^4",
         "symfony/console": "^2.8|^3.0|^4",
         "symfony/stopwatch": "^2.8|^3.0|^4",


saimaz:
Agree. In that case we will avoid the future conflicts. Thanks for the notice.

YetiCGN:
Thanks!

What about a merge to `master`? Is that a good idea? If so, should I open another PR or will you change it?

saimaz:
I will merge it to the master later. If you want to introduce any new change as a PR you should start from the branch as in the version you are targeting to release.

HcErwin:
I'm running v5.2.* for a while now on 7.4 locally and works great. Due to the restriction I can't update my test/staging server through my pipelines (can't hack like locally). Maybe we can unlock 7.4?

YetiCGN:
@HcErwin 6.0.* supports PHP 7.4, can you upgrade to the next major version?

HcErwin:
Sadly no, because our platform is quite big and is still locked to Elastic 5.*. Otherwise I would've certainly upgraded, but still discussing the upgrade to 6 or 7 for now :)

YetiCGN:
Then you could fork the repository and change the restriction in the ElasticsearchBundle `composer.json`, commit and push to your fork's branch and include that as a repository in your project's `composer.json` like this:

```
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/HcErwin/ElasticsearchBundle"
        }
    ],
```

HcErwin:
That sounds the most logical for now indeed; thank you! :D

patie:
can u please add support to PHP8? 🙏

PR #860, opened 2018-12-01
Fix php 7.2 compatibility for Elasticsearch 2

### LINKED ISSUE #828: `Object` is reserved word in PHP 7.2
Starting from PHP 7.2 we no longer can name our class `Object` which is still the case in ESB.

When trying to run project using ONGR on PHP 7.2 I get the following error: `PHP Fatal error:  Cannot use 'Object' as class name as it is reserved in ....vendor/ongr/elasticsearch-bundle/Annotation/Object.php on line 23`

https://github.com/ongr-io/ElasticsearchBundle/blob/bf49352edfef67dde4ce4084c2a921e59a8cf7db/Annotation/Object.php#L23

Consider renaming `Object` to something not reserved.

alexander-schranz:
We have to support still some elasticsearch 2 clusters in some of our projects, this PR should fix the php 7.2 compatibility based on makasim's work on the 5.x branch: https://github.com/ongr-io/ElasticsearchBundle/pull/852.

Fixes #828


saimaz:
I thought this version won't be used at all anywhere. Will merge and release. Have fun ;)

alexander-schranz:
@saimaz thank you for merge and release!

PR #868, opened 2019-03-18
A small upgrade to the v6

saimaz:


Annotation/Index.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    /**
+     * Index alias name. By default the index name will be created with the timestamp appended to the alias.
+     */
+    public $alias;
+
+    /**
+     * Index alias name. By default the index name will be created with the timestamp appended to the alias.
+     */
+    public $hosts = [
```
alexander-schranz: @saimaz Is this a good idea? Its not possible to use environment variables or parameters inside annotation or?

saimaz: There will be a config which can override index settings. The annotation settings were added to make things more simple. In that case, you won't need any config someone uses default settings (like the host is 127.0.0.1:9200 and index alias name is like the class name).

alexander-schranz: Ah ok, sounds good.

PS: the comment above $hosts looks false.

saimaz: This is why it is still a beta stage :) But if you could test this version and give some feedback I would appreciate it a lot.

alexander-schranz: I will try to test the version at end of the week.

saimaz: At the end of week, there will be docs written as wellr I think.

PR #869, opened 2019-03-21
[WIP] 6.0 major version release candidate

saimaz:
This is a work in progress version of release candidate version of the 6th major bundle version. There were many changes and improvements already done. Here's the list what is left to do:
- [x] Introduce configuration override layer (to let users use `.env` file to set index name and other parameters).
- [x] Finish functional tests
- [ ] Finish unit tests
- [ ] Introduce pipeline objective support
- [x] Update docs
- [ ] Add more test cases for MappingPass when there are no documents or the namespace is different
- [ ] Finish document generator with the `@Index` annotation support


ro0NL:
Hi @saimaz im really eager to test this bundle using Symfony 4 :) do you expect this to be possible any time soon?

saimaz:
My plan is to release it when I will finish adding more unit tests to cover at least 70% of the code and cover it with the docs. The progress is like 50% of both cases. The pipeline and other fancy support will add later on minor versions, don't want to stop the major release :). 

@ro0NL you may subscribe to get an email notification when there will be a release. It will definitely be during this month.

cadavre:
So... the month is now over for more than two months. :p I had to quit on Ongr already in one huge, long-term project. This means there will be no going back now. :/

Any updates on this?

einorler:
Great that there are already all the docs there. However, I do not agree that we should delete the doc folders from the bundles. I understand the reasoning behind putting docs on https://github.com/ongr-io/docs.ongr.io/, however, if we will remove the docs from the repositories of the bundles, we will lose the documentation for older versions of our bundles and only have the latest one available. I think that copy and paste would be a better solution than the deletion, with complete loss of the support for older version docs.

einorler:
Even though I admit that there are still quite a bit of test cases that could be covered and several features that need be implemented, this is already working and stable enough to release, a too long-awaited 6.0 version that the community has been asking for more than 9 months.  I'm releasing this as 6.0 and any additional features and more fine-tuned tests can be added as patches or minor releases as time moves on. Let's do this a step at a time, guys :)

alexander-schranz:
@saimaz it seems this is not released yet can we tag this with 5.2.5?

saimaz:
@alexander-schranz sure thing

alexander-schranz:
@saimaz Thank you! Can you ping me if its tagged?

saimaz:
@alexander-schranz [v5.2.5](https://github.com/ongr-io/ElasticsearchBundle/releases/tag/v5.2.5)

alexander-schranz:
@saimaz Thank you!

Issue #877, opened 2019-04-26
Possible optimisation in TerminateListener


esynaps:
Related to

https://github.com/ongr-io/ElasticsearchBundle/blob/c817dc53d3b93129e8575a740e11484c67f79a87/EventListener/TerminateListener.php#L50

This listener takes 300 ms to execute for me. The processing is long because I have entities with many properties (~ 100) and each property takes between 1ms and 3ms to be recovered by the doctrine (with annotations).

This listener is being processed at every request, even when no treatment with ES is to be done.
Why not add this check before recovering managers unnecessarily?

```
if(!$this->container->initialized(sprintf('es.manager.%s', $key))) {
  continue;
}
$manager = $this->container->get(sprintf('es.manager.%s', $key));
```

saimaz:
That makes sense, and in your case, it sounds like a significant issue regarding performance. Is there any chance you could make a PR to the 5.x branch?

esynaps:
I can propose a PR but before I wanted to make sure that this modification would have no undesirable effect.

saimaz:
The initialized check is entirely ok. Also in v6 of this bundle, this place will be different since there will be no more managers.

esynaps:
I can not run the tests locally. I am on Windows and I have a problem with EOL.

\ONGR\ElasticsearchBundle\Tests\Unit\Service\Json\JsonWriterTest::testPush
Failed asserting that two strings are equal

```
var_dump(str_replace(array( "\r"), 'R', str_replace(array("\n"), 'N', $expectedOutput)));
// string(93) "[RN{"count":2},RN{"_id":"doc1","title":"Document 1"},RN{"_id":"doc2","title":"Document 2"}RN]"
var_dump(str_replace(array( "\r"), 'R', str_replace(array("\n"), 'N', file_get_contents($filename))));
// string(89) "[N{"count":2},N{"_id":"doc1","title":"Document 1"},N{"_id":"doc2","title":"Document 2"}N]"
```

How can I avoid that?

esynaps:
The problem is still present and very restrictive. And considering the source code, it does not seem to have been fixed in 6.*.
On 500 ms to execute a request, I have 250ms used by the Terminate Listener of elasticsearch bundle whereas no request to send to the database is necessary.

PR #878, opened 2019-04-26
Add condition to ignore unused managers in TerminateListener

esynaps:
Related to https://github.com/ongr-io/ElasticsearchBundle/issues/877

### DIFF
--- a/EventListener/TerminateListener.php
+++ b/EventListener/TerminateListener.php
@@ -46,8 +46,15 @@ class TerminateListener
         foreach ($this->managers as $key => $value) {
             if ($value['force_commit']) {
                 try {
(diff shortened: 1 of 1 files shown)


esynaps:
The error in the build is unrelated to the PR : 

>  [ErrorException]          
>   Undefined index: version  

`The command "composer require --no-update symfony/symfony:${SYMFONY}" failed and exited with 1 during .`

alexandre-le-borgne:
Can we have news?

saimaz:
for 5.2 branch CI seems corrupted.

esynaps:
The problem is still present and very restrictive. And considering the source code, it does not seem to have been fixed in 6.*.

PR #879, opened 2019-05-28
Fix for memory problem in export index

PWalkow:
As file value was used to store all of the documents it stopped the script when the value was getting too big due to number of documents stored in it. Value was not used so I just removed it.

### DIFF
--- a/Service/ExportService.php
+++ b/Service/ExportService.php
@@ -82,7 +82,6 @@ class ExportService
         $filename = str_replace('.json', '', $filename);
         $writer = $this->getWriter($this->getFilePath($filename.'.json'), $metadata);
 
-        $file = [];
         foreach ($results as $data) {
             if ($counter >= $maxLinesInFile) {
                 $writer->finalize();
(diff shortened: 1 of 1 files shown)


PWalkow:
@saimaz As the CI have failed for entirely different reasons than the code change (as the value is not even used) I think this should be merged as it just breaks exporting big elasticsearch entries as it just depends on memory available to php due to growing value which is unused...

saimaz:
Don't worry about the CI fail. I'll take a look at it.

saimaz:
Btw, if you want to apply a patch to the 5.x version the right branch to open PR would be 5.2 instead of master. Master branch is upcoming 6.0 major release, and it won't be merged backward to 5.x branches.

PWalkow:
👍 , I've changed the base branch, I've used magic in-line-editing in github, I think it would pick up the version already that I was editing :)
is it all good to go now?

saimaz:
👍

PWalkow:
Any chances to have it merged soon?

carlcasbolt:
That's a nice little improvement on the memory usage - was that array intended for some debug at some point?

PWalkow:
@saimaz Any news on that, will it be merged anytime soon or there is still some issue with it?

PR #880, opened 2019-06-18
Configuration to allow request_options to be used

carlcasbolt:
I have found that I need to set some of the client request options for
the transport layer within the Connection via the Manager.

This change allows for any of the GuzzleHttp Client Request Options to
be set in yaml per index and used within the elastic search libraries.

- Carl.


PWalkow:
Thanks, we needed that as well!

carlcasbolt:
Have I missed something in this Pull Request? Travis is borking with `github-oauth.github.com is not defined.`

PR #882, opened 2019-07-25
Make mappings configuration unique to allow prepend  of configuration

alexander-schranz:
Sometimes you want to prepend a configuration in a bundle if then the developer of the bundle defines the same bundle the ongr elasticsearch bundle should not crash and just import it once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: false
+dist: trusty
 language: php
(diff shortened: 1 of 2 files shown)


saimaz:
Thanks for all contributions @alexander-schranz. I strongly suggest you upgrade to a newer version :)

PR #892, opened 2019-11-26
do not set _index in bulk data

wasinger:
in order to prevent error "explicit index in bulk is not allowed"
in ES instances where rest.action.multi.allow_explicit_index is set to false.

We don't need this _index setting in `bulk` method because in `commit`
method index is set to $this->getIndexName()


einorler:
Good catch this. In essence this was a relic from the earlier version where the ES manager was intended to work on multiple ES types. Thanks for the PR

PR #903, opened 2019-12-13
Only allow mapped fields to be serialized/unserialized

toooni:
There is currently an issue if there are additional getters inside the document which are not mapped. For example I am using some additional logic in a getter similar to this:
```
    /**
     * @ES\Embedded(class="App\Document\PackageVariation")
     */
    private $variations;

    public function getVariations(): Collection
    {
        return $this->variations;
    }

    public function setVariations(Collection $variations)
    {
        $this->variations = $variations;
    }

    public function getSpecialVariation(): ?PackageVariation
    {
        foreach ($this->variations as $variation) {
            if ($variation->isSpecial()) {
                return $variation;
            }
        }

        return null;
    }
```

This results in a document where `specialVariation` will be written to the index even if it isn't mapped.


toooni:
Since the symfony/serializer is not used anymore in 6.1+ I don't think it makes sense to merge this. Therefore I am closing the PR.

PR #909, opened 2020-02-02
Add support for es7 in version 5

alexander-schranz:
This change would allow to install the ES7 client and use it with the same api. It would help us that the community can use the newest ES version when using the SuluArticleBundle. Upgrading to the newest version of the bundle would be in our case to many bc breaks at current state.

The change itself is simple if its uses *Elasticsearch 7* the Index settings is set to:

```http
include_type_name=true
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,10 +22,10 @@
         "doctrine/cache": "~1.4",
         "doctrine/collections": "~1.4",
         "monolog/monolog": "~1.10",
-        "ongr/elasticsearch-dsl": "~5.0|~6.0"
+        "ongr/elasticsearch-dsl": "~5.0|~6.0|~7.0"
     },
     "require-dev": {
-        "elasticsearch/elasticsearch": "~5.0|~6.0",
+        "elasticsearch/elasticsearch": "~5.0|~6.0|~7.0",
         "mikey179/vfsstream": "~1.4",
         "phpunit/phpunit": "~5.6",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": ">=5.6,<7.4",
+        "php": "^5.6|^7.0",
```
alexander-schranz: for no warnings requires @toooni PR https://github.com/ongr-io/ElasticsearchBundle/pull/906

fredericbarthelet:
@alexander-schranz just noticed that I had to require `ocramius/package-version` in my project in order to ensure this package can retrieve `elasticsearch-dsl` version. Shouldn't this package be required in this project instead ?

alexander-schranz:
@fredericbarthelet don't wanted to break someones project by adding a additional package, but if @saimaz is ok for it we could add `ocramius/package-versions:^1.0` as requirement.

alexander-schranz:
@fredericbarthelet we decided to add Symfony 5 support to the v5 branch to fork this bundle. There the requirement for the ocramius/package-version is removed and is instead build on top of a [VERSION](https://github.com/elastic/elasticsearch-php/blob/80fe9b192f3806349df45a0e67749aabb625addc/src/Elasticsearch/Client.php#L35) constant by added php elasticsearch library.

The fork can be found here: https://github.com/handcraftedinthealps/elasticsearchbundle

Issue #902, opened 2019-12-12
Multiple indexes for one Document


toooni:
As I see it there is no chance to create more than one index using the same document type in version 6.0. Until now I have used my "Package" document in multiple indexes like "package_de_chf" or "package_fr_chf" (country/currency pairs). This was a performance gain since I never needed these entries to be mixed.

![image](https://user-images.githubusercontent.com/241080/70724104-4566fb00-1cfa-11ea-8b29-79a5d6c92544.png)
I've tried it like this. But obviously this won't work since only one `IndexService` is being generated with the name being the same as the classname.

Is is possible we could use the alias as service names? Maybe add an optional parameter `alias_as_service_name=true` for the annotation?

I am willing to create a PR if there is a chance it will get merged ;)

saimaz:
There is no way to do that with annotations but you can do that with a YAML config. If you won't find how will write an example later.

toooni:
I am not sure if you meant I could already do that or that it can be made to work with YAML config? As I have seen in `MappingPass`, the service name is fixed to the class name so I assume the latter.

BTW: It would also be possible with annotations if we introduce a new one:
```
/**
 * @ES\MultiIndex(indexes={
 *     @ES\Index(alias="package_de_chf")
 *     @ES\Index(alias="package_fr_chf")
 *     @ES\Index(alias="package_de_eur")
 *     @ES\Index(alias="package_fr_eur")
 * })
 */
```
When using this annotation we could automatically use the `alias` as the service name since it wouldn't be possible to use the class name anyway.

toooni:
@saimaz Can you elaborate? I am still not sure what's the case here.

toooni:
@saimaz I tried to figure it out myself again but I can't seem to find a solution since every `Namespace` can only be used once in the yaml config.

toooni:
@saimaz I just spent some time creating an alternative to the current config. I had an idea how we wouldn't loose BC:
```
ongr_elasticsearch:
    indexes:
        'package_de_chf':
            class: Weekend4two\BackendBundle\Document\Package
            hosts:
                - '%env(ELASTICSEARCH_ENDPOINT)%'
        'package_fr_chf':
            class: Weekend4two\BackendBundle\Document\Package
            hosts:
                - '%env(ELASTICSEARCH_ENDPOINT)%'
        Weekend4two\BackendBundle\Document\PackageIsochrone:
            hosts:
[...]

PR #917, opened 2020-02-11
[POC] Implented working config with multiple indexes per document class

toooni:
I've made a proof of concept for a solution where we can have multiple indexes for one document class again (https://github.com/ongr-io/ElasticsearchBundle/issues/902). The config can look like this:
```
ongr_elasticsearch:
    indexes:
        package_de_chf:
            class: App\Document\Package
            alias: package_de_chf
            hosts:
                - '%env(ELASTICSEARCH_ENDPOINT)%'
        package_fr_chf:
            class: App\Document\Package
            alias: package_fr_chf
            hosts:
                - '%env(ELASTICSEARCH_ENDPOINT)%'
        App\Document\PackageIsochrone:
            hosts:
                - '%env(ELASTICSEARCH_ENDPOINT)%'
```

If a `class` parameter is provided the `indexes` children keys can be a self defined string which will be used as the index service name (if `package_de_chf` is used, the service will be called `ongr.es.index.package_de_chf`).

I abused the `IndexSettings` class to create a temporary storage before creating the definition. I am not sure if this is a good solution or if we should just work with a separate class (only difference would be the `propertyMetadata`).

I've also removed the return values from the setters since those are not used anywhere.

This PR can probably be targeted to the 6.2 branch too.

### DIFF
--- a/DependencyInjection/Compiler/MappingPass.php
+++ b/DependencyInjection/Compiler/MappingPass.php
@@ -92,10 +92,11 @@ class MappingPass implements CompilerPassInterface
         foreach (array_diff($indexClasses, $overwrittenClasses) as $indexClass) {
             try {
                 $indexSettingsArray[$indexClass] = $this->parseIndexSettingsFromClass($parser, $indexClass);
-            } catch (DocumentIndexParserException $e) {}
+            } catch (DocumentIndexParserException $e) {
+            }
         }
 
-        foreach($indexSettingsArray as $indexSettings) {
+        foreach ($indexSettingsArray as $indexSettings) {
             $this->createIndex($container, $indexSettings);
         }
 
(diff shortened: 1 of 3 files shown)


toooni:
@saimaz Can I have an opinion on the proposed changes? I am not even sure this goes far enough. I think the new approach by using the class name as service name is generally flawed and you should rethink this design decision.
Also having the services public doesn't seem to follow best practices. IMO there should still be something like a manager container. The container would contain the lazy index services and returns them by name without the need of a prefix.

saimaz:
The current architecture does not support multiple indexes for the same document. I'm fully aware that the current approach is not seamless how OOP law directs. 

The main idea was to simplify as much as possible to work with the document as the index. For me, personally, the current approach is very handy.

Having said that I'm very open for suggestions, PR's or anything else that might help.

toooni:
Thank you for your response. Ok, with the approach I've implemented both will be possible. Can you review it? I'll create separate PR with my idea for the container approach later.
FYI: I've tested the proposed changes in this PR with our application @ weekend4two.com and had no issues (I've used both class names and manual names for the service names).

saimaz:
Sure, I'l take a look

toooni:
@saimaz Any idea when you will have a chance to look at this PR?

toooni:
Hi @saimaz. I just wanted to ask again if there is a chance to have this PR reviewed? Or if you don't want to allow the same document used in multiple services?

PR #921, opened 2020-02-20
Overriding index setting configuration bug

### LINKED ISSUE #920: Overriding index setting configuration bug
Hi guys,

It seems we got a bug at overriding the index settings configuration.
Here is how we can reproduce the issue.

Elasticsearch settings:
Screen Shot 2020-02-20 at 15 21 24

Document annotation:
Screen Shot 2020-02-20 at 15 26 46

Actually result:
Screen Shot 2020-02-20 at 15 21 09

Expected result:
`{
    "mappings": {
        "_doc": {
            "properties": []
        }
    },
    "settings": {
        "number_of_replicas":0,
        "number_of_shards": 5
    }
}`

I believe this [line](https://github.com/ongr-io/ElasticsearchBundle/blob/v6.2.1/DependencyInjection/Compiler/MappingPass.php#L78) which caused the bug, you can see the issue easily [here](http://sandbox.onlinephpfunctions.com/code/1ae50077e59cd700d591426ab1084d228d20e9b4)

    - PHP version: 7.3
    - Elasticsearch version: 6.6.1
    - Bundle version: 6.x
    - Symfony version: 5.x

petrmalina:
Fixes #920.

There were actually two problems.
One mentioned in the issue, merging values into arrays. That is fixed using array_replace_recursive instead of array_merge_recursive.
Second was caused by the array_filter and trying to set number_of_replicas to 0. Array filter converts the value to boolean FALSE and removes it from the array. I added a custom callback function to compare it using identical operator.


saimaz:
Thanks for the PR @petrmalina

saimaz:
Merged your commit in the https://github.com/ongr-io/ElasticsearchBundle/commit/769cd3d99034a38a5c21a477e7562d27b20313d9

PR #929, opened 2020-03-31
Remove managers from profiler view

### LINKED ISSUE #900: Profiler page doesn't seem to work
Opening the profiler panel doesn't seem to work. The following error is generated when clicking it inside the developer toolbar:

```
Neither the property "managers" nor one of the methods "managers()", "getmanagers()"/"ismanagers()"/"hasmanagers()" or "__call()" exist and have public access in class "ONGR\ElasticsearchBundle\Profiler\ElasticsearchProfiler".

in vendor/ongr/elasticsearch-bundle/Resources/views/Profiler/profiler.html.twig (line 122) 
```

I've updated to Symfony 4.4 from 4.3, so there shouldn't be any BC breaks, but I could be wrong and something may have changed.

The amount of requests show correctly in the developer toolbar, but clicking it results in the above error.

* PHP version: 7.3
* Elasticsearch version: 6.8.4
* Bundle version: v6.0.1
* Symfony version: 4.4.1

pkly:
It seems that managers were deleted in 7.x, but they're still used in this file which causes a crash when opening the profiler page for Elasticsearch

Fixes #900

### DIFF
--- a/Resources/views/Profiler/profiler.html.twig
+++ b/Resources/views/Profiler/profiler.html.twig
@@ -118,14 +118,6 @@
 {% block panel %}
     <h2>Queries</h2>
     {{ block('queries') }}
-    <h2>Managers</h2>
-    {% if collector.managers %}
-        {% include '@WebProfiler/Profiler/table.html.twig' with {data: collector.managers} only %}
-    {% else %}
-        <p>
-            <em>No managers.</em>
-        </p>
-    {% endif %}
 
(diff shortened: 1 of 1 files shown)


Koc:
Can someone merge this PR? I have same issue with profiler

// cc @saimaz

wizardz:
Can we please merge this ?

einorler:
This will be available with `v7.0.3`

PR #946, opened 2020-05-26
Fix es7 total hits

alexander-schranz:
Since https://github.com/ongr-io/ElasticsearchBundle/pull/909 the 5.2 version support ES 7 but there is false response of the count as the api will return total as object.


alexander-schranz:
@saimaz would be great if you could have a look at this.

alexander-schranz:
For all waiting for this we decided to create a fork of the `5.x` Version here: https://github.com/handcraftedinthealps/elasticsearchbundle

PR #972, opened 2020-12-07
6.2

einorler:


pkly:
It seems that php 7.1 support must be dropped. the rest is about using some const that doesn't exist @einorler

pkly:
Can some fix this up and please merge it to 7.x? @einorler @saimaz

pkly:
@einorler Could this possibly be merged already?

